PR #4, opened 2016-02-18
DNSimple - creating the same record twice gives an error

cdrx:
Creating the same record twice with DNSimple returns a HTTP 400 from the API which should be ignored


AnalogJ:
Hi @cdrx, thanks for the PR!
:+1: Looks good to me. I have to go through and add a bit more error handling for all the providers.

PR #8, opened 2016-03-14
Support all CloudFlare's accepted record types

welwood08:
As per https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record there are a few more record types available from this provider, can't think of a reason to exclude them (acceptable lexicon options should include all types that are acceptable to at least one provider).


AnalogJ:
:+1 Thanks for the PR. It'll be included in the next release.

PR #50, opened 2016-07-30
make sure we can specify the TTL for a provider.

### LINKED ISSUE #49: TTLs hard-coded and not even passed to providers from the command line
I just started to use lexicon with acme.sh and after banging my head to the wall for a while (and even suspecting that DNS Made Easy's API was not working properly as my records kept being generated with TTL of 84600 no matter what I did) before I realized to check the sources here - and, lo and behold, for example here..

https://github.com/AnalogJ/lexicon/blob/067f7feecaf2b79222374804f02b03ecdd4aa347/lexicon/providers/dnsmadeeasy.py#L37

..the value of TTL really is hard-coded to 84600 (value varies in other providers).

_(Not that I should need to explain my frustration, but that isn't very good for example for TXT records that are only very infrequently queried (so short TTL does not matter) but in the case of errors really do need to be able to be fixed quicker than that as processing the certificates via ACME protocol is highly automated and things can go wrong sometimes - mostly because of user errors.)_

Also, the value from `--ttl` is never even passed to the providers (that also are not prepared to receive it) when creating or updating the record, as seen here:

https://github.com/AnalogJ/lexicon/blob/067f7feecaf2b79222374804f02b03ecdd4aa347/lexicon/client.py#L28

..and here:

https://github.com/AnalogJ/lexicon/blob/067f7feecaf2b79222374804f02b03ecdd4aa347/lexicon/providers/dnsmadeeasy.py#L32

This seems like a pretty elementary bug to me, maybe somebody lost their train of thought while implementing this parameter? :)

AnalogJ:
This PR fixes #49 and ensures that we can correctly specify the TTL for a record when creating/updating it. Sets a default TTL of 3600s (60 minutes) which is standardized across all providers and more reasonable than 86400 which was hardcoded in some providers previously

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ Potential providers are as follows. If you would like to contribute one, please 
 - AuroraDNS ([docs](https://www.pcextreme.com/aurora/dns))
 - AHNames ([docs](https://ahnames.com/en/resellers?tab=2))
 - AWS Route53 ([docs](https://docs.aws.amazon.com/Route53/latest/APIReference/Welcome.html))
-- BuddyDNS ([docs](https://www.buddyns.com/support/api/v2/))
-- ClouDNS ([docs](https://www.cloudns.net/wiki/article/56/))
-- DurableDNS ([docs](https://durabledns.com/wiki/doku.php/ddns))
+- ~~BuddyDNS ([docs](https://www.buddyns.com/support/api/v2/))~~ 
+- ~~ClouDNS ([docs](https://www.cloudns.net/wiki/article/56/))~~ Requires paid account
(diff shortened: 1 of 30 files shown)


CapsuleCD:
Hi.

I'm an automated pull request bot named [CapsuleCD](http://www.github.com/AnalogJ/capsulecd). I handle testing, versioning and package releases for this project. 

[CapsuleCD](http://www.capsulecd.com/#!/deploy/github/AnalogJ/lexicon/pullrequests/50)

---
If you're interested in learning more about [CapsuleCD](http://www.github.com/AnalogJ/capsulecd), or adding it to your project, you can check it out [here](http://www.github.com/AnalogJ/capsulecd)

Issue #35, opened 2016-04-21
Digitalocean Provider spits out invalid json


napcae:
```
[  
   {  
      'content':u'"v=DKIM1; k=rsa; p=MIGfMA0GC[...redacted...]GUrwFVgQIDAQAB"',
      'id':1256011,
      'type':u'TXT',
      'name':'mesmtp._domainkey.ctnguyen.net',
      'ttl':''
   },
   {  
      'content':      u'"v=spf1 include:spf.messagingengine.com -all"',
      'id':2349116,
      'type':u'TXT',
      'name':'@.ctnguyen.net',
      'ttl':''
   }
]
```

The issued command is ```lexicon digitalocean list ctnguyen.net --auth-token $LEXICON_DIGITALOCEAN_TOKEN TXT```

Look after the `content`-pair and `type`-pair. There are _u_'s after the colon. I don't were there are coming from. 
I discovered this, when I tried to setup letsencrypt.sh with the provided example hook. 
The `clean_challenge` returns an empty list(maybe because the json is invalid).

napcae:
What exactly do you need me to do? Regarding the test suites?

```bash
# lexicon digitalocean delete ctnguyen.net --name '_acme.challenge.ctnguyen.net' --auth-token $LEXICON_DIGITALOCEAN_TOKEN TXT
Namespace(action='delete', auth_token='32dc[...redacted...]d559348', content=None, domain='ctnguy
en.net', identifier=None, name='_acme.challenge.ctnguyen.net', priority=None, provider_name='digitalocean', ttl=None, type='TXT')
list_records: []
[]
Traceback (most recent call last):
  File "/usr/local/bin/lexicon", line 9, in <module>
    load_entry_point('dns-lexicon==1.1.2', 'console_scripts', 'lexicon')()
  File "/usr/local/lib/python2.7/dist-packages/lexicon/__main__.py", line 51, in main
    client.execute()
  File "/usr/local/lib/python2.7/dist-packages/lexicon/client.py", line 37, in execute
    return self.provider.delete_record(self.options.get('identifier'), self.options.get('type'), self.options.get('name'), self.options.get('content'))
  File "/usr/local/lib/python2.7/dist-packages/lexicon/providers/digitalocean.py", line 88, in delete_record
    raise StandardError('Record identifier could not be found.')
StandardError: Record identifier could not be found.
```
I meant that list, see line 5?

Regarding the --auth-token:
````
[...]

PR #70, opened 2016-10-27
Support pagination from DigitalOcean API 

mrwacky42:
Fixes #35. 
This is branched from #69, so this will be much more readable after that lands.

PR #73, opened 2016-11-17
Support paginated records from Digital Ocean

mrwacky42:
Supercedes #70.

It's been a while, but I think this is the correct patch minus PEP8.


mrwacky42:
Fixes #35 without PEP8.

AnalogJ:
@mrwacky42 Awesome, now to figure out why my packaging bot didn't notice this PR 😞 

I should have this merged sometime today, thanks for the PR!

PR #82, opened 2016-12-28
Added check against subdomain for Route53 zones

michaeltravisuk:


AnalogJ:
Hey @michaeltravisuk 
Support for delegated subdomains has been added using the `--delegated` flag. 
Can you confirm that it fixes the problem you were trying to solve with this PR?

PR #83, opened 2017-01-02
Yandex PDD provider

kharkevich:
Hi
I create support Yandex PDD as a DNS provider.
But, I skip test because I haven't ideas to implementation it.
@AnalogJ  please merge if it possible.


paskal:
Thanks for this PR! This provider is called yandex and not yandex_pdd, but now (thankfully) we have [Yandex.Cloud DNS](https://cloud.yandex.com/en/docs/dns/), I guess I'll give it a shot writing a provider for it. I'll set the name to yandex_cloud unless objections.

Issue #87, opened 2017-01-06
luadns provider script breaks with default ttl


jmbromley:
When leaving ttl unspecified, the script ends up calling the luadns API with the ttl parameter set as 'None' instead of the default.  This results in the call failing.  I'm creating a pull request with a fix for this.

stompro:
Same problem with dnsmadeeasy provider.

AnalogJ:
This should be fixed in #92 for all providers.

Can you try using version `v1.2.1`?

stompro:
I just tested 1.2.1 on a different vm and it works with no modifications, thanks.
Josh

PR #88, opened 2017-01-06
Fix use of default ttl posting 'None' as ttl (issue #87).

jmbromley:
Fixes issue #87

### DIFF
--- a/lexicon/providers/luadns.py
+++ b/lexicon/providers/luadns.py
@@ -27,7 +27,7 @@ class Provider(BaseProvider):
 
     # Create record. If record already exists with the same content, do nothing'
     def create_record(self, type, name, content):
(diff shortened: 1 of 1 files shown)


stompro:
This seems to be a problem for other providers also... the construct "self.options.get('ttl',self.default_ttl)" doesn't work as intended.  It returns none instead of the default_ttl.

self.options.get('ttl') or self.default_ttl} seems to work for dnsmadeeasy also.

AnalogJ:
Yeah, I'll merge this and take a look at the other providers. Thanks for the PR!

AnalogJ:
Thanks for the PR!
I fixed this issue in #92 for all providers.

Can you try using version `v1.2.1`?

PR #90, opened 2017-01-08
Fix default ttl settings for dnsmadeeasy

stompro:
The code to use the default ttl didn't work, it would result in a value of "none" instead of 3600.  This change needs to be made elsewhere also.
Signed Off: Josh Stompro <github@stompro.org>


AnalogJ:
Thanks for the PR!
I fixed this issue in #92 for all providers.

Can you try using version `v1.2.1`?

PR #104, opened 2017-03-13
Add dnsimple provider's V2 API as new provider -- V1 is deprecated.

nealian:
From the [v1 documentation](https://developer.dnsimple.com/v1/):

> DNSimple API v1 is deprecated and in maintenance-only. API v1 will be discontinued on October 1st, 2017.

This adds API v2 as a new provider, though it should probably honestly be a modification to the current... I don't know how you intend to handle different API versions, though, so I figured that this was a good starting point.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/deploy/github/AnalogJ/lexicon/pullrequests/104)

---

AnalogJ:
Hey @nealian 
Thanks for the PR! Yeah, I would prefer to just replace the current DNSimple provider with your version. 

I did notice that this would be a breaking change, as your updated provider doesn't support email & password authentication like the old one did. If you could make that change, that would make the migration even easier. 

In any case I'll do a minor version bump when I merge your PR. 

Thanks again!

nealian:
My pleasure!

That was part of the reason I didn't want to use the same name, actually...
Unfortunately, DNSimple's API v2 doesn't support username+password
authentication or the v1 account/domain tokens, it only supports the new
app-specific access tokens for users or the associated accounts. In
addition, all of the API endpoints are different (almost all starting with
/<account_number> now, and the records in /<account>/zones/<domain>/records
instead of /domains/<domain>/records), with different methods (PATCH vs
PUT, for example, for updating a record)... It really is a wholly different
beast.

I suppose I could merge this into the existing DNSimple provider, but it
would definitely be a kludge.

Thoughts?

lexicon/providers/dnsimplev2.py
```diff
@@ -109,7 +110,6 @@ def update_record(self, identifier, type=None, name=None, content=None):
     def delete_record(self, identifier=None, type=None, name=None, content=None):
         if not identifier:
             records = self.list_records(type, name, content)
-            print(records)
```
nealian: Had this here for debugging, and forgot to remove it

AnalogJ:
Hey @nealian 

Forgive my ignorance, I was basing the support for HTTP Basic Auth on a quick look at their [v2 docs](https://developer.dnsimple.com/v2/#authentication). Are you saying that a Basic Auth doesn't provide the permissions needed to update the records, and you need to go through the OAuth flow?
Or am I just completely mis-reading how they use Basic Auth with their API?

Yeah, I agree. Mixing and matching v2 and v1 in the same provider doesn't make sense. And since v1 is explicitly marked as deprecated, its probably a good idea to just completely replace the existing provider. I'm was thinking that a minor version bump would be enough, but I did do a major bump when I removed providers previously, so I think that might be a good practice to follow in this case as well.

nealian:
Oh hey, I'm happy to see that you are completely correct and I simply missed the HTTP Basic Auth mechanism in my late-night sprint.  I'll add that mechanism back straight-away!

nealian:
I've added username/password auth (with optional 2FA) and otherwise replaced dnsimple with this.

AnalogJ:
🎉  Awesome. I'll review it and if everything looks good I'll create a new release.

AnalogJ:
Thanks @nealian , I've released a new version of `lexicon` with your PR.

weppos:
I was doing some due diligence, and I just discovered that a PR was provided to upgrade lexicon to API v2.

Thank you very much @nealian, highly appreciated! Thanks also @AnalogJ for your support.

PR #107, opened 2017-03-28
Added support for provider Glesys.

hecd:


lexicon/providers/glesys.py
```diff
@@ -0,0 +1,131 @@
...
+class Provider(BaseProvider):
+
+    def __init__(self, options, provider_options={}):
+        super(Provider, self).__init__(options)
+        self.domain_id = None
+        self.api_endpoint = provider_options.get('api_endpoint') or 'https://api.glesys.com'
+
+    # Provider does not support this. Uses basic auth.
+    def authenticate(self):
+        pass
```
AnalogJ: The authenticate method is primarily used to validate that the domain whose DNS records we're going to update/modify is actually owned by the user. Something like the https://github.com/glesys/API/wiki/functions_domain#domainlist endpointshould be used.

The idea is to fail quickly if something is wrong.

AnalogJ:
Hey @hecd thanks for the PR!

I've made a comment about the `authenticate()` method. If you can fix that, I think this PR is good to merge :)

hecd:
Thanks for the feedback!

I've implemented authenticate() as suggested. :)

AnalogJ:
This was merged in v2.0.2

Thanks again 🥇

PR #118, opened 2017-04-12
Update did not actually find the identifier correctly if not passed in specifically

tnwhitwell:
I missed this in testing - it will now find the identifier when updating, as it does when deleting.
I'm so sorry about the commits here. I've gotten myself in a pickle.


AnalogJ:
LGTM, though not sure why circleci tests didnt run.

AnalogJ:
Looks like it passed the CapsuleCD tests and merged, so 👍 

If you could add a test to the integration suite that covers this case, that would be awesome. Just so no one else gets bitten by this. 

Either way, thanks for the PR!

PR #129, opened 2017-06-14
Include OVH provider

adferrand:
This pull request makes Lexicon able to operate on a DNS domain managed by OVH.

OVH is the leading dedicated server, cloud solution and web hosting in Europe, and top 5 in North America. So I think it worths integrating OVH in Lexicon.

To manipulate a DNS zone, OVH provides an API which needs an *Application Key*, *Application Secret*, *Consumer Key* and *Entrypoint*. They form together the token to obtain access on the API for a given account. On CLI, for the OVH provider, following optional parameters are available:
* `--auth-application-key`
* `--auth-application-secret`
* `--auth-consumer-key`
* `--auth-entrypoint`

Theses parameters may also be setted with environment variables `LEXICON_OVH_APPLICATION_KEY`, `LEXICON_OVH_APPLICATION_SECRET`, `LEXICON_OVH_CONSUMER_KEY` and `LEXICON_OVH_ENTRYPOINT`.

Text helper is provided on the CLI to explain this, and gives the right URL to request a correctly configured token from OVH (see `lexicon ovh --help`).

This pull request has been successfullly tested, with correctly configured env variables:
* on CLI (eg. `lexicon ovh create elogium.net TXT --name hello --content world`),
* on all integration tests.

Please note that I deviated a little from the standard build process described in *CONTRIBUTE.md*: I made an override of *_test_options* in *test_ovh.py* to call *env_auth_options*, because auto-fill of auth parameters by env variable was not working without this modification.

I am available to any further discussions in order to integrate this pull request.

Regards,
Adrien Ferrand

PR #130, opened 2017-06-14
Repair and reintegrate Gandi provider

adferrand:
Gandi provider was removed from Lexicon because of incompatibilites on some Python version (see commits 319ac2a46 and bf8ca76df61).

I do not have the details, but I assume that the problem was the xmlrpc python client, which is named xmlrpclib for Python2.X and xmlrpc.client for Python3.X.

This PR solves the issue by making a try/except on the xmlrpc module import : *gandi.py* will try first to import *xmlrpclib* module (which will work on Python2.X), and in case of *ImportException* will import *xmlrpc.client* module aliased as *xmlrpclib* (which will work on Python3.X).

Some incompatibilites with recent versions of Lexicon are also solved, including the ability of *update_record* to update a DNS record without identifier, but by matching on name and content, and providing a default TTL of 3600s.

Similarly to #129, *_test_options* has been overriden to use *env_auth_options* and consequently allows use of environment variables during tests.

I am available to further discussions in order to integrate this PR.

Regards,
Adrien Ferrand


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/130)

---

lexicon/providers/gandi.py
```diff
@@ -0,0 +1,277 @@
...
+                pass
+
+            finally:
+                if not ret and version is not None:
+                    self.api.domain.zone.version.delete(self.apikey, self.zone_id, version)
+
+        print("delete_record: {0}".format(ret))
+        return ret
+
+    def _fqdn(self, name):
```
AnalogJ: This doesn't look necessary, there's already a `_fqdn_name` function available in the superclass. https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/base.py#L84

lexicon/providers/gandi.py
```diff
@@ -0,0 +1,277 @@
...
+        if not name.endswith('.') and not name.endswith('.{0}'.format(self.domain)):
+            name += '.{0}'.format(self.domain)
+        return name
+
+    def _canonicalize_name(self, name):
```
AnalogJ: I think this can also be removed and replaced with a call to `_relative_name` from the superclass. https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/base.py#L98

lexicon/providers/gandi.py
```diff
@@ -0,0 +1,277 @@
...
+                                             'ttl': self.options.get('ttl') or self.default_ttl
+                                            })
+            self.api.domain.zone.version.set(self.apikey, self.zone_id, version)
+            ret = True
+
+        finally:
+            if not ret and version is not None:
+                self.api.domain.zone.version.delete(self.apikey, self.zone_id, version)
+
+        print("create_record: {0}".format(ret))
```
AnalogJ: can you replace all print statements with `logger.debug` usage? https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/cloudflare.py#L75 is a good example of how that's done in other providers.

tests/providers/test_gandi.py
```diff
@@ -0,0 +1,24 @@
...
+# the tests which *each and every* implementation of the interface must
+# pass, by inheritance from define_tests.TheTests
+class GandiProviderTests(TestCase, IntegrationTests):
+
+    Provider = Provider
+    provider_name = 'gandi'
+    domain = 'reachlike.ca'
+
+    def _test_options(self):
+        cmd_options = env_auth_options(self.provider_name)
```
AnalogJ: env_auth_options don't really make sense on a build server. You should be able to load the default test_options and then add any overrides you need. https://github.com/AnalogJ/lexicon/blob/82fa5056df2122357af7f9bec94aebc58b247f91/tests/providers/test_powerdns.py#L18

adferrand:
Ok, modifications are pushed.

AnalogJ:
🎉 Thanks again!

adferrand:
You're welcome ;)

Issue #151, opened 2017-11-24
nsone  -- list index out of range.


init-js:
using dns-lexicon 2.1.11

I'm getting a list index out of range in the process of cleaning up acme/certbot TXT records on nsone.

```
/opt/certbot # lexicon nsone list sitelabs.io TXT --auth-token MYTOKEN
Arguments: Namespace(action='list', auth_token='MYTOKEN', content=None, delegated=None, domain='sitelabs.io', identifier=None, name=None, priority=None, provider_name='nsone', ttl=None, type='TXT')
Starting new HTTPS connection (1): api.nsone.net
https://api.nsone.net:443 "GET /v1/zones/sitelabs.io HTTP/1.1" 200 None
Starting new HTTPS connection (1): api.nsone.net
https://api.nsone.net:443 "GET /v1/zones/sitelabs.io HTTP/1.1" 200 None
Traceback (most recent call last):
  File "/usr/local/bin/lexicon", line 11, in <module>
    load_entry_point('dns-lexicon==2.1.11', 'console_scripts', 'lexicon')()
  File "/usr/local/lib/python2.7/site-packages/lexicon/__main__.py", line 69, in main
    client.execute()
  File "/usr/local/lib/python2.7/site-packages/lexicon/client.py", line 44, in execute
    return self.provider.list_records(self.options.get('type'), self.options.get('name'), self.options.get('content'))
  File "/usr/local/lib/python2.7/site-packages/lexicon/providers/nsone.py", line 68, in list_records
    'content': record['short_answers'][0],
IndexError: list index out of range
```

The current record list for TXT records is (from the nsone panel):

![image](https://user-images.githubusercontent.com/1110751/33194468-0735c3b6-d085-11e7-91a2-f99b7b658957.png)

Has there been an nsone response format change?

init-js:
Adding more info.

Dumping the record information before processing reveals that the problem is with nsone "linked" records.

```
...
{u'domain': u'sitelabs.io', u'short_answers': [u'dns1.p03.nsone.net', u'dns2.p03.nsone.net', u'dns3.p03.nsone.net', u'dns4.p03.nsone.net'], u'ttl': 3600, u'tier': 1, u'type': u'NS', u'id': u'599ce1d3a632f60001a95c76'}
[...]

PR #157, opened 2017-12-06
Fixing dnsone driver to support linked records

init-js:
When a linked record is encountered, this patch resolves the link (recursively, up to 3 times). Fixes issue #151.

This would be further improved with caching of answers. I suspect it is common to have multiple records in a zone linking to the same record. Definitely an improvement, regardless.

Also, the use of "type" as a keyword parameter throughout the module's code shadows the builtin "type", but I left it as it was. I suspect this requires a deeper refactoring.

### DIFF
--- a/lexicon/providers/nsone.py
+++ b/lexicon/providers/nsone.py
@@ -52,31 +52,90 @@ class Provider(BaseProvider):
         logger.debug('create_record: %s', 'id' in payload)
         return 'id' in payload
 
+    def _find_record(self, domain, _type=None):
+        """search for a record on NS1 across zones. returns None if not found."""
+
+        def _is_matching(record):
+            """filter function for records"""
+
+            if domain and record.get('domain', None) != domain:
+                return False
+            if _type and record.get('type', None) != _type:
+                return False
+            return True
+
+        payload = self._get('/search?q={0}&type=record'.format(domain))
+        for record in payload:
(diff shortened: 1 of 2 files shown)


init-js:
As an extra note, each link record encountered during a zone listing (assumed to be possibly useful to answer the lexicon query), is looked up with an additional API call.

There is a slight complication in that the target of the linked record could be anywhere on NS1 (e.g. foo.example.org can link to bar.example2.com), and the actual target zone can't be determined from the metadata in the source record -- it is likely that the target zone is going to be the tld, which could be extracted by stripping all subdomains, but the NS1 zone could be any subdomain prefix. 

Therefore, the extra lookup is currently implemented as a search API call ( https://ns1.com/api#search-zones-and-records ), followed by a filter in a for-loop.  If the target zone was known ahead of time, then the more precise GET https://api.nsone.net/v1/zones/:zone/:domain/:type API endpoint could be used instead.

init-js:
great.

Would you like me to create a separate issue to add tests for creating a few linked records on ns1, and testing that they are returned correctly? I've done the manual testing only, and I don't think the current tests have that coverage.

AnalogJ:
Yeah, if you could add a separate PR for that work, that would be great. 

I'm also going to have to figure out if there's a better way to support provider specific functionality. 

Thanks again for your PR!

PR #177, opened 2018-02-27
Fix spelling

louis-lau:
Every fix counts!


AnalogJ:
Haha fair enough :) thanks for the PR!

PR #189, opened 2018-03-16
Rackspace provider handle duplicate record on create

rmarscher:
I was hitting duplicate record exceptions using the lexicon hook with dehydrated to generate and verify challenge responses. This update will have it retry with an update call if the create record call fails due to there being a duplicate. There was also a bug with the content filter for list_records. Rackspace API uses a query param named data instead of content for that filter.


rmarscher:
Closing this for now. I'm getting a 503 error from Rackspace using the data filter for the [search records API](https://developer.rackspace.com/docs/cloud-dns/v1/api-reference/records/#search-records). I'm unsure what is causing the 503 error, but it seems likely that this fix could introduce another problem.

PR #190, opened 2018-03-19
Test suite fixes for #182, Record set support & duplicate record noop testing. 

AnalogJ:
Still work in progress. 

The providers which need recordings are as follows:

- [x] aurora - contributed by @joostdebruijn
- [x] cloudflare
- [ ] cloudns - contributed by @ppmathis
- [x] cloudxns
- [x] digitalocean
- [x] dnsimple
- [x] dnsmadeeasy
- [ ] dnspark
- [ ] dnspod
- [ ] easydns
- [ ] gandi - contributed by @hrunting
- [ ] gehirn - contributed by @chibiegg
- [ ] glesys - contributed by @hecd
- [ ] godaddy - contributed by @adferrand
- [x] linode
- [x] luadns
- [x] memset - contributed by @tnwhitwell
- [ ] namecheap - contributed by @pschmitt
- [x] namesilo
- [x] nsone - **paid account - need contributor**
- [ ] ovh - contributed by @adferrand
- [x] pointhq
- [ ] powerdns - contributed by @insertjokehere
- [ ] rackspace - contributed by @rmarscher
- [x] rage4
- [ ] route53 - contributed by @eadmundo
- [ ] sakuracloud - contributed by @chibiegg 
- [ ] softlayer - contributed by @adherzog 
- [ ] transip - contributed by @LordGaav @yorickvP
- [ ] vultr - **paid account - need contributor**
- [x] yandex
- [ ] zomomi - contributed by @jarossi 

fixes #182 
fixes #138 

---
If you are working on adding recordings for the new tests, please do the following:

- checkout the `record_set_support` branch from AnalogJ/lexicon
- then run `py.test tests/providers/test_memset.py` to generate the updated recordings. Read `CONTRIBUTING.md` for getting your dev environment setup. 
- you may need to update the provider to add support for record sets before the tests pass correctly.
- add any new files created in the `tests/fixtures/cassettes/{PROVIDER_NAME}/` directory. 
- open a PR against `record_set_support`

I wrote a cleanup script in tests/cleanup.sh that you can use to remove any existing DNS records in your account (just change the Provider name).

PR #203, opened 2018-03-26
Part 2: Test suite fixes for #182, Record set support & duplicate record noop testing

### LINKED ISSUE #182: Support multiple records of the same type on an identifier
In [certbot](https://github.com/certbot/certbot), we use lexicon to create TXT records for a domain to prove to a SSL/TLS certificate authority that the user has control over that domain so the CA will give you a certificate which you can use to set up HTTPS. Due to the way [the protocol](https://tools.ietf.org/html/draft-ietf-acme-acme-09) works, we sometimes have to set multiple TXT records for a single domain.

I'm not sure if each of the providers we use have problems with creating multiple TXT records on a domain (although I know this works with your code for dnsimple), but they all have problems deleting it and immediately error out if there is more than 1 record for the domain:

* [cloudxns](https://github.com/AnalogJ/lexicon/blob/a8c478ad44c06012c3aaaac077f70e67c12a136e/lexicon/providers/cloudxns.py#L125)
* [dnsimple](https://github.com/AnalogJ/lexicon/blob/903af58378ab9942d817c57e0330b5f7ac26b4e9/lexicon/providers/dnsimple.py#L128)
* [dnsmadeeasy](https://github.com/AnalogJ/lexicon/blob/9ad8a9cbf9e4b5c24fe525b274966c2432baf132/lexicon/providers/dnsmadeeasy.py#L121)
* [luadns](https://github.com/AnalogJ/lexicon/blob/a8c478ad44c06012c3aaaac077f70e67c12a136e/lexicon/providers/luadns.py#L98)
* [nsone](https://github.com/AnalogJ/lexicon/blob/8b649338b6c03c7bfb2d465249cd759de62296cb/lexicon/providers/nsone.py#L175)

Can support for this be added?

### LINKED ISSUE #200: Problem with update_record api
This is not an issue per se. Rather, a design flaw in lexicon.
This issue is not present in providers that have some form of concrete record identifier.
I'll try to explain it.
Almost all providers provide support for multiple records with the same name and type. For example, I can have a record named "test" of type "A" and value "1.1.1.1" and "2.2.2.2". In zonefile, it would look something like this:
```
test 3600 IN A 1.1.1.1
test 3600 IN A 2.2.2.2
```
On providers that have record identifier implemented, I can target either one of these and change it. For instance, I can change "2.2.2.2" to "3.3.3.3" and it would look like this:
```
test 3600 IN A 1.1.1.1
test 3600 IN A 3.3.3.3
```

The problem begins when a provider does not implement record ids. Without a record id, there is currently no way to target only one of the records.
The current signature of `update_record` is `update_record(identifier, type, name, content)`.
The parameter `content` is the new content the needs to be placed there. There is no option to specify the existing content. This effectively means that it's impossible to update records like these without a concrete identifier implemented by the provider.
Under the current implementation of most lexicon providers, the first record returned by the provider will be updated. Under most cases, this means that calling `update_record(None, 'A', 'test', '3.3.3.3')` will update `1.1.1.1` to `3.3.3.3`. But this is unpredictable because if the provider were to reorder the result in a list call, `2.2.2.2` could get updated instead.

I understand that this problem is difficult to fix in the current major version of the lexicon with so many other stuff depending on it but this problem should be fixed asap.

### LINKED ISSUE #201: Extend the test suite
We need to add the following tests:

- #200 - update specific record in record set (using query, not id) 
- auto pagination?
- #191  - TTL tests should be enabled by default (only disabled on rare providers who do not support it at all).

AnalogJ:
Follow up to PR #190 

Removed all commented out tests & suites that should be enabled, with tests. 

The providers which need recordings are as follows:

- [x] aurora
- [ ] cloudflare
- [x] cloudns
- [ ] cloudxns
- [ ] digitalocean
- [ ] dnsimple
- [ ] dnsmadeeasy
- [ ] dnspark
- [ ] dnspod
- [ ] easydns
- [ ] gandi
- [ ] gehirn
- [ ] glesys
- [ ] godaddy
- [x] linode
- [ ] luadns
- [ ] memset
- [x] namecheap
- [ ] namesilo
- [ ] nsone
- [ ] ovh
- [ ] pointhq
- [ ] powerdns
- [x] rackspace
- [ ] rage4
- [ ] route53
- [ ] sakuracloud
- [ ] softlayer
- [x] transip
- [ ] vultr
- [ ] yandex
- [ ] zomomi

TODO List:

- [ ] Add tests to ensure that update works correctly with record sets
- [ ] Add tests to ensure that setting TTL works correctly when updating and deleting record. 
- [ ] Create recordings for any providers I have access to. 
- [x] use CODEOWNERS to ping all developers who have access to paid providers, and ask them to create recordings & make fixes so that test pass. 

---
If you are working on adding recordings for the new tests, please do the following:

- checkout the `record_set_support_contrib` branch from AnalogJ/lexicon
- then run `py.test tests/providers/test_{PROVIDER_NAME}.py` to generate the updated recordings. Read `CONTRIBUTING.md` for getting your dev environment setup. 
- you may need to update the provider to add support for record sets before the tests pass correctly.
- add any new files created in the `tests/fixtures/cassettes/{PROVIDER_NAME}/` directory to git. 
- open a PR against `record_set_support_contrib`

I wrote a cleanup script in `tests/cleanup.sh` that you can use to remove any existing DNS records in your account (just change the provider and domain name).


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/203)

---

joostdebruijn:
@AnalogJ I see that Aurora still passes all tests, is it still necessary to submit new recordings?

Yesterday I also had a look at the TransIP provider (which I also use), I can submit new recordings for that. There was one test failing, I'll have look at it tonight.

AnalogJ:
Breaking this up into multiple PR's/Issues because I've taken too long to get this PR merged.

PR #205, opened 2018-03-26
multiple updates for Namecheap client

jvanasco:
This proposed PR handles the following issues:
* #191
* #172 

# Overview

* the namecheap `authenticate` method now calls the cloud endpoint `namecheap.domains.getInfo`, which will show information for a 'permissioned' domain or return an error document. Previously this called the `domains_getList` client method and iterated paginated results to find the domain, however that list only shows domains "owned" by a namecheap account, and does not include domains owned by a first account which have had permissions delegated to the second active account.

* the namecheap provider did not respect the TTL commandline arg. The new behavior is to inject a TTL from the commandline if provided and valid, otherwise leaving the TTL value blank (which will invoke the serverside default). Submitting a ttl outside the valid range for namecheap will raise a `ValueError`.

* the IntegrationTests object was extended with a `provider_variant` attribute, defaulted to `None`.  If present, this will change the cassette path from `{provider}/IntegrationTests` to `{provider}/{variant_name}-IntegrationTests`.  This approach allows for testing multiple domain scenarios for the same provider.

* the `NamecheapProviderTests` replaced a `domain` attribute with a `domain` property tied to the environment variable `LEXICON_NAMECHEAP_DOMAIN`. This will allow developers to test against the llive API without adjusting code.

* A variant of `NamecheapProviderTests` called `NamecheapManagedProviderTests` was added, which tests against a domain that is not owned, but managed, on the namecheap platform.

* The namecheap tests were documented with instructions on how to fully setup the test environment and create sandbox accounts in the cloud.

Issue #191, opened 2018-03-19
Default TTLs don't seem to be implemented


jvanasco:
I noticed this while working a PR for issue #172

When creating a record on Namecheap, no TTL is set:

    https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/namecheap.py#L65

Looking at a few other providers this seems consistent.  There is an explicit `--ttl` flag, it's just not used.

It might make sense to default TTLs to something short (60s) and then allow the --ttl flag to override that.  With a lot of DNS systems, leaving it null will engage a default value on the serverside that could be from 30mins to 24hours, which can be a mess to work around if it gets cached.

AnalogJ:
ah, thats definitely wrong. a TTL value passed in via `--ttl` should definitely be processed by lexicon.

You are correct, it looks like `namecheap` does not handle that option correctly, while other providers like `cloudflare` do.  

I initially wanted to have a test suite that implemented a TTL test, however I ended up having to remove it due to the fact that the range of acceptable TTL values varied significantly across providers. 

Once I finish work on #190, I'll come back and take a look at this. In the meantime if you could update the `namecheap` provider as part of your PR that would be awesome!

jvanasco:
I'll include the namecheap fix in my PR

I went through the providers, and these seem to not implement the TTL option.  I might have missed one or two:

* digitalocean
* linode (hardcodes 0)
* namecheap
* nsone
* pointhq

trinopoty:
Linode has a ttl option called 'Default' and a ttl value of 0 tells the api to apply this 'Default' ttl.

trinopoty:
Also, the range of ttl values supported by different providers varies greatly and may change over time. This factor needs to be considered too.

jvanasco:
> Linode has a ttl option called 'Default' and a ttl value of 0 tells the api to apply this 'Default' ttl.

The current provider doesn't accept the `--TTL` option from the commandline though.  The `0` is hardcoded into the provider and the default is managed on the server.  Using the `0` as a fallback default seems like the right choice, but the commandline flag in `options` should override it.

jvanasco:
@AnalogJ one quick question - is there a way in your tests setup that I can only one provider at a time?  i can't seem to figure out how.

AnalogJ:
Hey @jvanasco 
Yeah you can do it by running the following:

`py.test tests/providers/test_{PROVIDER_NAME}.py`

[...]

Issue #172, opened 2018-02-15
namecheap authentication fails for permissioned domain management


jvanasco:
The following lines check to see if the domain name is returned by Namecheap's API

https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/namecheap.py#L54-L63

Namecheap's API seems to be halfassed and this won't work as expected.

please bear with me on this, it's going to be odd to explain.

The only domains that seem to appear in this list are ones which are OWNED by the authenticating account. The list does not include items that are MANAGED by the authenticating account.  Namecheap's backend provisioning systems use an ACL (Access Control List) that allows for an domain "Owner" to proxy/delegate different responsibilities for a domain to other users of their system.

In my case, "Account1" is the `owner` of all domains, while "Account2" has been delegated admin privileges for **only** the DNS settings of a subset of domains.  This was done to allow a tier of company staff to manage DNS via the control panel AND also enable API access -- but keeping the staff and API away from registration/transfer/etc operations.

Privileged domains don't seem to appear on this paginated view and will trigger lexicon exiting.

A potential check would be the `namecheap.domains.getinfo` endpoint (https://www.namecheap.com/support/api/methods/domains/get-info.aspx)

Supplying a `&DomainName=example.com` argument will result in a valid response if the user has permissions:

```
<ApiResponse Status="OK" xmlns="http://api.namecheap.com/xml.response">
  <Errors />
  <Warnings />
  <RequestedCommand>namecheap.domains.getinfo</RequestedCommand>
  <CommandResponse Type="namecheap.domains.getInfo">
    <DomainGetInfoResult Status="Ok" ID="11111111111" DomainName="EXAMPLE.COM" OwnerName="USERNAME" IsOwner="false" IsPremium="false">
      <DomainDetails>
        <CreatedDate>02/15/2000</CreatedDate>
        <ExpiredDate>02/15/3000</ExpiredDate>
        <NumYears>0</NumYears>
      </DomainDetails>
      <LockDetails />
      <Whoisguard Enabled="False">
        <ID>0</ID>
      </Whoisguard>
      <PremiumDnsSubscription>
        <UseAutoRenew>false</UseAutoRenew>
        <SubscriptionId>-1</SubscriptionId>
        <CreatedDate>0001-01-01T00:00:00</CreatedDate>
        <ExpirationDate>0001-01-01T00:00:00</ExpirationDate>
        <IsActive>false</IsActive>
      </PremiumDnsSubscription>
[...]

PR #162, opened 2018-01-20
add support for Gandi LiveDNS API

sumpfralle:
Gandi currently provides the old (xmlrpc-based) and the new (REST-like) API for accessing the domain.

PR #207, opened 2018-03-26
multiple updates for Namecheap client

jvanasco:
This replaces PR #205

This proposed PR handles the following issues:
* #191
* #172 

# Overview

* the namecheap `authenticate` method now calls the cloud endpoint `namecheap.domains.getInfo`, which will show information for a 'permissioned' domain or return an error document. Previously this called the `domains_getList` client method and iterated paginated results to find the domain, however that list only shows domains "owned" by a namecheap account, and does not include domains owned by a first account which have had permissions delegated to the second active account.

* the namecheap provider did not respect the TTL commandline arg. The new behavior is to inject a TTL from the commandline if provided and valid, otherwise leaving the TTL value blank (which will invoke the serverside default). Submitting a ttl outside the valid range for namecheap will raise a `ValueError`.

* the IntegrationTests object was extended with a `provider_variant` attribute, defaulted to `None`.  If present, this will change the cassette path from `{provider}/IntegrationTests` to `{provider}/{variant_name}-IntegrationTests`.  This approach allows for testing multiple domain scenarios for the same provider.

* the `NamecheapProviderTests` replaced a `domain` attribute with a `domain` property tied to the environment variable `LEXICON_NAMECHEAP_DOMAIN`. This will allow developers to test against the llive API without adjusting code.

* A variant of `NamecheapProviderTests` called `NamecheapManagedProviderTests` was added, which tests against a domain that is not owned, but managed, on the namecheap platform.

* The namecheap tests were documented with instructions on how to fully setup the test environment and create sandbox accounts in the cloud.

### DIFF
--- a/lexicon/providers/namecheap.py
+++ b/lexicon/providers/namecheap.py
@@ -45,6 +45,20 @@ def ProviderParser(subparser):
 
 
 class Provider(BaseProvider):
+    """
+    ========================================================================
+    WARNING
+    ========================================================================
+    The Namecheap API does not add/update/delete a host but "gets" and
+    "sets" ALL hosts at once (a complete replacement). In their comments on
+    the API docs (https://www.namecheap.com/support/api/methods/domains-dns/set-hosts.aspx)
+    it appears that not all host types are handled by their system.
+
+    Their API only handles `[A, AAAA, CNAME, MX, MXE, TXT, URL, URL301, FRAME]`
+
+    If you have SRV record, it may get lost. Also records configured as
+    `A + DDNS` on their control panel will be downgrated to `A` records.
+    """
 
     def __init__(self, options, engine_overrides=None):
         super(Provider, self).__init__(options, engine_overrides)
@@ -55,7 +69,7 @@ class Provider(BaseProvider):
             UserName=options.get('auth_username',''),
             ClientIP=options.get('auth_client_ip',''),
             sandbox=options.get('auth_sandbox', False),
-            debug=False
+            debug=False,
         )
         self.domain = self.options['domain']
         self.domain_id = None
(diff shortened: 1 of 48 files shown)


jvanasco:
Somehow I missed pushing a commit before the Pull Request and the default domain did not stick. That has been fixed, and namecheap tests all pass now:

> tests/providers/test_namecheap.py ...............s.ss........................s.ss.........

AnalogJ:
Hey @jvanasco 
Awesome looking PR, let me give some feedback item-by-item. 

- `namecheap.domains.getInfo` looks exactly like what we want, if testing for a specific domain instead of looking up all domains and filtering on the client side is perfect. 
- ensuring that TTL values specified by the client are passed into the API requests is exactly what I want to see. However I'm unsure if we want to handle provider limits on the client side at this time. I'd rather let the provider specify that a TTL is invalid, rather than specifying them in lexicon and having the provider change them out from under us. 
- `provider_variant` is perfect. I came up with something similar during the discussion in #162. Your general purpose solution will work there as well. 
- `domain` being a property that can be set via env variables is interesting. I want to think about it some more, but making it a general purpose pattern for all provider test suites may be possible without too much work. 
- `NamecheapManagedProviderTests` is great
- love the documentation!


So the TLDR for getting this PR merged is: for now, can you remove the TTL validation code. I think there needs to be a bigger discussion about doing client side validation as a whole (TTL, Record Types, character sets, record lengths , etc) first.

jvanasco:
sounds good. I'll get you a PR tomorrow!

jvanasco:
I pulled out the TTL validation. I left it as a function though, because the docstring was worth preserving and didn't fit well in the 'create' 'update' or __main__ sections.

in terms of the `domain` being controlled by the environment, I didn't mention this on the other thread, but you'd be able to manage your own copies of domains+credentials in a blackbox (https://github.com/StackExchange/blackbox), ansible `vault` or other cryptographic store.

AnalogJ:
Thanks, this is great!

Issue #204, opened 2018-03-26
A list of issues present in lexicon


trinopoty:
I'll create a list of issues that I've come across.
- Issue with update_record mentioned in #200 
- No TTL implementation in some providers
- No method to specify accepted value of TTL by different providers
- No policy to handle illegal TTL value in providers
- Unpredictable provider behavior in `create_record`, `list_record`, `update_record` and `delete_record`
- No easier way to specify advanced options in case of MX, SRV, etc records

# TTL Issue
Providers should be able to tell the user about the range/list of TTL values accepted by it.
A policy should be defined on how to handle invalid TTL values supplied by the user.

# Unpredictable behavior
Different provider implementations behave differently in certain situations at the moment. The behavior for the methods `create_record`, `list_record`, `update_record` and `delete_record` should be specified for most/all possible situations.
This includes record sets and the like.

# Easier reading of advanced options
This included TTL and extra options that MX,SRV, etc records require. Instead of `self.options.get`, we could try passing these in an argument if possible.

I'll update this as I come across other issues.

AnalogJ:
this is great, I started something like this in #201 but I'll probably close that out in favor of this issue.

If you could, could you include links/numbers for any existing issues in your description?

And yes please, I'd like to get a comprehensive list of provider inconsistencies before I start extending the test suite again and contacting developers to make recordings. Updates would be appreciated!

trinopoty:
I was thinking more along the lines of matching the provided ttl to the closest one supported by the provider. For instance, godaddy takes a minimum of 600, so providing anything less, the provider will use 600. Something like that.
Let's not have the user worry about exceptions on invalid values.
Additionally, provide a list/range of accepted ttl values per provider so the user can choose between those.

trinopoty:
@AnalogJ is it possible to change method signatures or do we have to go with the existing signatures?

AnalogJ:
@jvanasco I originally looked at wrapping all exceptions by a set of lexicon specific ones, however I decided against it at the time for 3 reasons: 

- while there's a finite amount of exception types that we might handle, there's incredible diversity in how they are generated by providers. Also, some providers do not distinguish between various error types, they just return a single error code. 
- most people will be using lexicon as a CLI, where exception types are not as useful
- laziness :) - it was alot of work, and I didnt have time to handle all the error cases for all the various providers I was adding 

[...]

PR #211, opened 2018-03-29
DRAFT: Added provider specification

trinopoty:
This is related to #204 
Please ignore my non-existent writing skills.
Suggest changes to this here.

### DIFF
--- a/SPECIFICATION.md
+++ b/SPECIFICATION.md
@@ -52,24 +52,38 @@ and `create_record('example.com','A','127.0.0.2')` is called, modify to this:
 
 ## list_record
 Read all records within a set and return them as individual records.
+
 If service returns this:
+
 `{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1', '127.0.0.2' ]}`
+
 Provider should return this:
+
 `[{'name':'example.com', 'type':'A', 'content': '127.0.0.1'},
 {'name':'example.com', 'type':'A', 'content': '127.0.0.2'}]`
 
 ## update_record
 Update only the record that matches the filter content.
+
 If current value is this:
+
 `{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1', '127.0.0.2' ]}`
+
 and `update_record('example.com','A','127.0.0.1','127.1.0.1')` is called, modify to this:
+
 `{'name':'example.com', 'type':'A', 'records':[ '127.1.0.1', '127.0.0.2' ]}`
+
 **Do not modify/remove other records in the set**
 
 ## delete_record
 Remove only the record that matches the filters.
+
 If current value is this:
+
 `{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1', '127.0.0.2' ]}`
+
 and `delete_record('example.com','A','127.0.0.1')` is called, modify to this:
+
 `{'name':'example.com', 'type':'A', 'records':[ '127.0.0.2' ]}`
+
 **Remove the record set only if `content` is not specified or length of `records` become 0 after removing record**
--- a/SPECIFICATION.md
+++ b/SPECIFICATION.md
@@ -1,0 +1,75 @@
+<h1>Provider Specification</h1>
+
+[TOC]
+
+# General
+- **name** Any record name supplied to the provider must be the full hostname.
+- **ttl** Reasonable default is 6 hours since it's supported by most services. Any service that does not support this must be explicitly mentioned somewhere.
+
+# API Operations
+## create_record
+- **Normal Behavior** Create a new DNS record.
+- **If Record Already Exists** Do nothing. **DO NOT** throw exception.
+- **TTL** If not specified or set to `0`, use reasonable default.
+- **Record Sets** If service has provisions for record sets, **DO NOT** remove any existing records in the set.
+
+## list_record
+- **Normal Behaviour** List all records. Apply filter as required.
+- **Record Sets** Read all record sets and add a record entry as follows:  {name:'', type:'', content: ''}. **DO NOT** bundle together record sets and treat as one record.
+- **Linked Records** For services that support some form of linked record, do not resolve, treat as CNAME.
+
+## update_record
+- **Normal Behaviour** Update a record matching name, type and content.
+- **Record Sets** If matched record is part of a record set, only update the record that matches. **DO NOT** modify other records in any way. Update the record set so that records other than the matched one are unmodified.
+- **TTL**
+    - If not specified, do not modify ttl.
+    - If set to `0`, reset to reasonable default.
+- **No Match** Throw exception?
+
+## delete_record
+- **Normal Behaviour** Remove a record from the service.
+- **Record sets** Remove only the record that matches all the filters. 
+    - If content is not specified, remove the record set.
+    - If length of record set becomes 0 after removing record, remove the record set.
+- **No Match** Do nothing. **DO NOT** throw exception
+
+# Record Set Handling (Clarification)
+Normally, a record set is formed from the combination of the name and type fields (on most services). Extra clarifications for handling these are given below.
+
+## create_record
+### No record set exists with the same name and type
+Create a record set with 1 record.
+`{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1' ]}`
+
+### Record set exists with same name and type
+Add the content of the new record to the record set.
+If service returns this:
+`{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1' ]}`
+and `create_record('example.com','A','127.0.0.2')` is called, modify to this:
+`{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1', '127.0.0.2' ]}`
+
+**NOTE** Existing data not modified
+
+## list_record
+Read all records within a set and return them as individual records.
+If service returns this:
+`{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1', '127.0.0.2' ]}`
+Provider should return this:
+`[{'name':'example.com', 'type':'A', 'content': '127.0.0.1'},
+{'name':'example.com', 'type':'A', 'content': '127.0.0.2'}]`
+
+## update_record
+Update only the record that matches the filter content.
+If current value is this:
+`{'name':'example.com', 'type':'A', 'records':[ '127.0.0.1', '127.0.0.2' ]}`
+and `update_record('example.com','A','127.0.0.1','127.1.0.1')` is called, modify to this:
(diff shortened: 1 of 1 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/211)

---

AnalogJ:
This is great. Once we flesh out the spec, we'll need to add the appropriate tests. 

I'll take my first pass at reviewing the spec now.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
+<h1>Provider Specification</h1>
+
+[TOC]
+
+# General
+- **name** Any record name supplied to the provider must be the full hostname.
```
AnalogJ: I'm not sure this line makes sense. Different providers have different requirements, some need FQDN, some support full name and relative names. We can't control the API's interface. 

On the client input/CLI side though, I've been focusing on being as flexible as possible. Given a FDQN, full name or relative name, lexicon should do the right thing. It's why I have tests like these: https://github.com/AnalogJ/lexicon/blob/master/tests/providers/integration_tests.py#L75-L91

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+- **ttl** Reasonable default is 6 hours since it's supported by most services. Any service that does not support this must be explicitly mentioned somewhere.
```
AnalogJ: Adding validation to lexicon has a couple of advantages:

- better usability, users know immediately if something is wrong
- better error messages

But there are also some disadvantages:

- our specified provider TTL range may be out-of-date with the actual API, will require a new version to be released. 
- having a lexicon specific TTL default breaks the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Now instead of letting the API determine the default TTL if none is provided, 
lexicon will send its own default, which may not be expected. 
- if we decide to go down the route of client validation, we should also consider provider specific validation for record types, record content length, record character sets, and more.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+
+# API Operations
+## create_record
+- **Normal Behavior** Create a new DNS record.
+- **If Record Already Exists** Do nothing. **DO NOT** throw exception.
+- **TTL** If not specified or set to `0`, use reasonable default.
+- **Record Sets** If service has provisions for record sets, **DO NOT** remove any existing records in the set.
```
AnalogJ: I think this is supposed to say: 

**Record Sets** If service supports record sets, **DO NOT** create a new records in the set.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+
+## list_record
+- **Normal Behaviour** List all records. Apply filter as required.
```
AnalogJ: I think this should be:

- **Normal Behaviour** List all records. If filters are provided, send to the API if possible, else apply filter on the client side. Return value should be a list of records.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+
+[TOC]
+
+# General
+- **name** Any record name supplied to the provider must be the full hostname.
+- **ttl** Reasonable default is 6 hours since it's supported by most services. Any service that does not support this must be explicitly mentioned somewhere.
+
+# API Operations
+## create_record
+- **Normal Behavior** Create a new DNS record.
```
AnalogJ: Should be changed to:
- **Normal Behavior** Create a new DNS record. Return a boolean `True` if successful.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
+<h1>Provider Specification</h1>
+
+[TOC]
+
+# General
+- **name** Any record name supplied to the provider must be the full hostname.
+- **ttl** Reasonable default is 6 hours since it's supported by most services. Any service that does not support this must be explicitly mentioned somewhere.
+
```
AnalogJ: We should add a general definition for a lexicon record here:

- **record** All provider/API records will be translated into a common lexicon record format. This format has the following structure:

```
{
    'type': string, // upper case, valid record type. eg. A, CNAME, TXT
    'name': string, // lowercase, full domain name. eg. test.record.example.com
    'ttl': integer, // positive integer, in seconds. eg. 3600
    'content': string,
    'id': string // optional, provider specified unique id. 
}
```

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+# API Operations
+## create_record
+- **Normal Behavior** Create a new DNS record.
+- **If Record Already Exists** Do nothing. **DO NOT** throw exception.
+- **TTL** If not specified or set to `0`, use reasonable default.
+- **Record Sets** If service has provisions for record sets, **DO NOT** remove any existing records in the set.
+
+## list_record
+- **Normal Behaviour** List all records. Apply filter as required.
+- **Record Sets** Read all record sets and add a record entry as follows:  {name:'', type:'', content: ''}. **DO NOT** bundle together record sets and treat as one record.
```
AnalogJ: not sure if this line is necessary, with `record` definition above.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+- **Linked Records** For services that support some form of linked record, do not resolve, treat as CNAME.
```
AnalogJ: I like this, will be interesting to come up with a valid test for it though.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+
+## update_record
+- **Normal Behaviour** Update a record matching name, type and content.
```
AnalogJ: May need to change the text:

- **Normal Behaviour** Update a record. Record to be updated can be specified by providing id `OR` name, type and content. Only content can be changed, not TTL or type. Return value will be a boolean `True`

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+- **Record Sets** If matched record is part of a record set, only update the record that matches. **DO NOT** modify other records in any way. Update the record set so that records other than the matched one are unmodified.
+- **TTL**
```
AnalogJ: I don't think we should provide a way to modify TTL, only content. 
Unfortunately TTL is not supported with all our providers, and if it is, this can be done by deleting a record and re-creating with a new TTL. 

I'm open to feedback here though.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+    - If not specified, do not modify ttl.
+    - If set to `0`, reset to reasonable default.
+- **No Match** Throw exception?
```
AnalogJ: *No Match* is an interesting case. Hmm, right now this is completely dependant on how the provider api handles it. If we provide an id that doesn't exist, I'd assume the provider will return an error, however if the user provides name, type and content, the record is found locally first, before attempting to update on the provider.

I think you may be right, the only consistent action here is to throw an exception.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+
+## delete_record
+- **Normal Behaviour** Remove a record from the service.
```
AnalogJ: - **Normal Behaviour** Remove a record or record set. Record to be deleted can be specified by providing id OR name, type and content. Record sets can be removed by providing name and type. Return value is a boolean `True`

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+- **Record sets** Remove only the record that matches all the filters. 
+    - If content is not specified, remove the record set.
+    - If length of record set becomes 0 after removing record, remove the record set.
+- **No Match** Do nothing. **DO NOT** throw exception
```
AnalogJ: This has the same problem as the update command with **No Match**
The way that we handle records vs the API's do is always going to be different. 

We'd have to handle the following via tests

-  delete with valid parameters - returns true
- delete with invalid id - returns false (ignore exceptions from API)
- delete with invalid filters - returns false
- delete that throws an error - dont ignore the exception from API -- not sure what this test would actually look like.

SPECIFICATION.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Record Set Handling (Clarification)
```
AnalogJ: Hmm, I'm not totally clear on the content below.

I dont think I want to change the record structure that lexicon generates. these examples also dont take into account record id's provided by some services.

SPECIFICATION.md
trinopoty: Requiring that the client provides FQDN and pushing the job of handling the provider requirements to the provider will make for a more consistent API. The user will not have to know whether to provide FQDN or relative based on different providers.

SPECIFICATION.md
trinopoty: I figured 6 hours is a reasonable value supported by most providers. If it does not make sense, we can just have the provider handle it internally. Though, having a single default for all providers will make the API more consistent.

SPECIFICATION.md
trinopoty: See clarification in the section below.

SPECIFICATION.md
trinopoty: I figured since linked records are not supported by most providers, treating it as CNAME will make more sense since it behaves that way.

SPECIFICATION.md
trinopoty: I think an update should be able to modify TTL. Then again, deleting and recreating also makes sense. Both are equally valid options.

SPECIFICATION.md
trinopoty: I'm not suggesting any changes to record structures. Those are just clarifications on handling of record sets.
They need to be merged into the main api sections later.

trinopoty:
If we do add an `options` parameter, we can set a few different options for controlling the strictness of various operations. For instance, we can add an option for controlling whether the API should ignore deletion of non-existent record. Or, add an option to make TTL value strict and throw exception if provider does not support it.

trinopoty:
I'll update some parts according to suggestions later today.

alexzorin:
I may have missed it elsewhere, but what is the standard for what record types need to be supported? Should providers give some indication of what types they support, or just throw `ValueErrors` at runtime for those they don't? Does that belong here at all?

trinopoty:
We have to define a few different exceptions for handling all the possible error cases.

AnalogJ:
I think this is a good first draft of the spec. I'll merge it as is right now. 

I've added a couple new entries to #204 from this discussion:

- we do not specify which Record Types are supported
- Should lexicon wrap/provide comprehensive list of custom error types for error handling downstream? 
- SPEC should specify that records should be autopaginated when possible.

trinopoty:
Should have waited a little longer and completed the specs.

Issue #222, opened 2018-04-23
Name conflict with lexicon


DimStar77:
There is a confusion/conflict with the pypi package lexicon.

lexicon (https://pypi.org/project/lexicon/) and dns-lextcon (this source) both install as 'lexicon' on the system, thus the 2 can't coexist on the system (they overwrite each other\s __init__.py)

trinopoty:
This can't be fixed without a new major version as far as I can tell.

jvanasco:
i'm +1 on renaming the library's install path to `dns_lexicon` to respect the PyPi name and referring to the project as "DNS Lexicon".

AnalogJ:
Is there a way to do this without affecting users who are already using `lexicon` as a library?

adferrand:
Developpers will need ultimately to update their code. But you can ease the migration path by exposing a dummy lexicon module, which delegates everything to the dns_lexicon module. Then deprecates lexicon module and put warnings when it is used to advertise users that lexicon namespace will be removed soon.

PR #228, opened 2018-05-09
[WIP] Update to new specification

trinopoty:
**DO NOT MERGE**
Updating lexicon to new specification.


trinopoty:
Don't mind the branch name. I couldn't come up with anything better.

trinopoty:
@AnalogJ have a look and update me.

AnalogJ:
I think i see where you're going with this, but the PR is a bit hard to follow with the provider changes, test suite changes and recording changes all together

I'll try to give some feedback in line.

trinopoty:
There isn't a whole lot of changes if you ignore the recordings. Basically, I tweaked the test suite a little, added some tests and updated the linode providers accordingly.

trinopoty:
Added exceptions and changed the method signatures of `create_record` and `update_record`.

AnalogJ:
Yeah, its the `action_options` I wanted to comment on, but I got a bit distracted. 

Whats' the point of passing that in? Within the provider, `self.options` already contains the cli options merged with the defaults and env overrides

https://github.com/AnalogJ/lexicon/blob/master/lexicon/client.py#L31

I know we talked about thread safety in another issue/PR, but that's a larger issue, and it'll require more changes than just this.

trinopoty:
Although the `self.options` contains the cli options, having an independent options parameter for the CRUD methods helps with isolation. Also, as far as thread safety goes, with this implementation, one provider object can serve operations on one domain across any number of threads.
One can, for instance, authenticate and then call any method from any thread without any side effects. Passing values in `self.options` interfere with isolation and thread safety.
I was also considering passing MX options and SRV options in the `action_options` parameter.

trinopoty:
If this PR is released as a new version (I think it will be required because of changes in method signatures), it'll be possible to update the install path to `dns_lexicon` to resolve #222 .

trinopoty:
I think it conforms to the specs pretty well.
Awaiting comments.

trinopoty:
@oldium I didn't merge the env options into the `options` argument because they are already in the `self.options` variable. Since we're only providing the auth data using env vars, and auth data is global for a provider instance, it didn't make much sense to put it in the `options` parameter which is supposed to be specific for a call.
About the `priority` parameter, I plan to add it to the `options` argument.

oldium:
I understand the reasoning behind, I am just not sure about the actual fix. Is there a reason to have the `options` as a dictionary instead of real parameters? What I missed during implementation of Subreg API provider were `ttl` and `priority` as direct parameters to `create_record` and `update_record` methods. So the other way around could be:

```
def create_record(self, type, name, content, ttl=None, priority=None):
    # Use ttl and priority directly, do not even ask self.options
```
Is there anything more to create a record?

Just an unrelated note - the Subreg SOAP API does neither need `ttl`, nor `priority` (they could be present or missing - server uses its own defaults), so even supplying defaults to those values does not make sense (like the `ttl` value `3600` in the current `base.py` code).

With direct `ttl`/`priority` arguments the caller of lexicon API (`client.py` for example) would be responsible to correctly pass the `ttl` and `priority` arguments, taking them for example from command line or environment variables (for compatibility reasons).

trinopoty:
It's implemented as a dictionary because different record types require different parameters, for instance, MX requires priority, SRV requires priority, weight, port and target. It'll not be a good design to add all of this as parameters. Implementing as a dictionary also ensure future proofing in case someone comes up with a new record type that requires some other parameters.

oldium:
SRV in Subreg API expects weight, port and target in the `content` value (separated by spaces), so supplying them separately in the `options` argument would not help much. If you think about supplying those parameters separately, this cannot be done without proper abstraction (or a lot of if-elses in each provider implementation). You may be looking for something like this (this is what Subreg API would need, used visitor design pattern in the `to_api` method):

```
class record(object):
    def __init__(self):
        self.type = None
    def to_api(self, provider):
        raise Exception("Not implementex)

class srv_record(record):
    def __init__(self, service, protocol, domain, ttl, priority, weight, port, target):
        super(self, record).__init__()
        self.type = "SRV"
        ... store arguments ...

    def to_api(self, provider):
        return provider.srv_to_api(self.service, self.protocol, self.domain, self.ttl, self.priority, self.weight, self.port, self.target)

class SomeProvider(Provider):
    def srv_to_api(self, service, protocol, domain, ttl, priority, weight, port, target):
        return "SRV", "{}.{}.{}".format(service, protocol, domain), "{} {} {}".format(weight, port, target), ttl, priority

    def create_record(self, record):
        if record.type == "MX":
            ... do specific things for MX for example if to_api call shown below does not suffice ...
        else:
            type, name, content, ttl, priority = record.to_api(self)
            ... call provider web API ...

    def parse_from_api(self, type, name, content, ttl, priority):
        if type == "SRV":
            return srv_record(... fill with decoded name, content, ttl and priority ...)
        elif type == "MX":
            return mx_record(...)
        else:
            ...
```

So I might ask - what is the real goal of extending the `create_record` method call by a generic `options` dictionary? Is it just supplying `ttl` and `priority`, or even more details, which are specific to record types? For the former case adding two arguments would be better, for the latter case having particular classes for each type would be better. But the latter case is completely different API than what the lexicon currently has.

trinopoty:
At present, it's the later case. @AnalogJ what's your view on creating classes for different record types?

trinopoty:
So, what's the verdict? Do we move towards classes for records?
________________________________

oldium:
> The only reason Im hesitant is that I want to keep the provider & CLI syntax as backward compatible as possible, especially since lexicon is used by certbot.

There is still a chance to keep the backwards compatibility - replace old implementation by a simple classes, which parse the `content` based on the `type` and create a record (helper/factory class will be useful - to not repeat parsing in all providers) and call new implementation. All new implementation will either live in the same package (`lexicon.providers`) with new name (naming convention needed), or in a new package. This way you will have a transition period, where old implementation could be replaced one-by-one until everything is migrated.

trinopoty:
Keeping the CLI backward compatible is not so difficult, the API though, that's a problem.

trinopoty:
That's a whole lot of extra code and effort and I'm not sure it's worth the effort. I'd suggest moving to the next major version with the new API. The current version will retain all existing providers and the API. All new providers will implement the new API. Any existing users of the library will continue to work. New users will use the new API. That's the best path in my opinion.
________________________________

oldium:
I also think that cleanest is to move to new major version. The old version could still be supported for some time and the _certbot_ could use a more strict version dependency in its `setup.py` like
```
dns-lexicon>=2.2.1, <3
```
until it migrates to new version.

AnalogJ:
Yeah, doing this work in a major release sounds like the way to go then. 
The "class per record type" pattern seems to be the way to go.

trinopoty:
I'll just close this PR and work on a new one then. This will require updates to the specs too. I'll work on that along side.

PR #239, opened 2018-06-06
provider: online

kapouer:
Hi, this a `online` provider (a.k.a. online.net, scaleway, dedibox...).

Feel free to comment while i still care !

### DIFF
--- a/tests/providers/test_online.py
+++ b/tests/providers/test_online.py
@@ -1,0 +1,26 @@
+# Test for one implementation of the interface
+from lexicon.providers.online import Provider
+from integration_tests import IntegrationTests
+from unittest import TestCase
+import pytest
+
+# Hook into testing framework by inheriting unittest.TestCase and reuse
+# the tests which *each and every* implementation of the interface must
+# pass, by inheritance from integration_tests.IntegrationTests
+class OnlineProviderTests(TestCase, IntegrationTests):
+
+	Provider = Provider
+	provider_name = 'online'
+	domain = 'capsulecd.com'
+
+	def _filter_headers(self):
+		return ['Authorization', 'x-recruitment']
+
+	def _test_engine_overrides(self):
+		overrides = super(OnlineProviderTests, self)._test_engine_overrides()
+		overrides['fallbackFn'] = (lambda x: 'placeholder_' + x if x != 'priority' else '')
+		return overrides
+
+	@pytest.mark.skip(reason="manipulating records by id is not supported")
+	def test_Provider_when_calling_delete_record_by_identifier_should_remove_record(self):
+		return
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -35,6 +35,7 @@ lexicon/providers/namecheap.py		@pschmitt @rbelnap
 lexicon/providers/namesilo.py		@analogj
 lexicon/providers/nsone.py			@init-js @trinopoty
 lexicon/providers/onapp.py			@alexzorin
+lexicon/providers/online.py     @kapouer
 lexicon/providers/ovh.py			@adferrand
 lexicon/providers/pointhq.py		@analogj
 lexicon/providers/powerdns.py		@insertjokehere
--- a/lexicon/providers/online.py
+++ b/lexicon/providers/online.py
@@ -1,0 +1,269 @@
+from __future__ import absolute_import
+from __future__ import print_function
+
+import json
+import logging
+
+import requests
+
+from .base import Provider as BaseProvider
+
+logger = logging.getLogger(__name__)
+
+
+def ProviderParser(subparser):
+    subparser.add_argument("--auth-token", help="specify private api token")
+
+def to_data(type, content):
+    if type == "TXT":
+        return '"{0}"'.format(content)
+    else:
+        return content
+
+class Provider(BaseProvider):
+
+    def __init__(self, options, engine_overrides=None):
+        super(Provider, self).__init__(options, engine_overrides)
+        self.zone_name = 'Zone Automatic Lexicon '
+        self.domain_id = None
+        self.passive_zone = None
+        self.active_zone = None
+        self.api_endpoint = self.engine_overrides.get('api_endpoint', 'https://api.online.net/api/v1')
+
+    def authenticate(self):
+        payload = self._get('/domain/')
+        domain = self.options['domain']
+        did = None
+        for row in payload:
+            if row['name'] == domain:
+                did = row['id']
+                break
+
+        if did is None:
+            raise Exception('No domain found')
+
+        self.domain_id = did
+        self.init_zones()
+
+    def list_zones(self):
+        return self._get('/domain/{0}/version'.format(self.domain_id))
+
+    def init_zones(self):
+        # sets current zone version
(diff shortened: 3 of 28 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/239)

---

kapouer:
Here it is !
(Squashed in a single commit)
I did not understand (not being a python dev) how to run the tests (only with that provider), but i manually tested obvious things like create/update/delete A or TXT.

AnalogJ:
Hey @kapouer 
Thanks for the PR! 

Looks like the only failure is in the transip provider, which @adferrand  just submitted a fix for. 
Lets get your PR retested and verify that its working correctly.

AnalogJ:
I dont see any test recordings, so you'll need to commit those if possible. 
The instructions for how to make recordings are included in the `Contributing.md` file.

CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/239)

---

AnalogJ:
I also noticed that you're missing a test configuration file, and API recordings so that the `Online` provider will be automatically tested by the test suite.

Without those, I will be unable to merge this PR. 

The instructions are available in the CONTRIBUTING.md file. These two sections will probably be the most valuable:
- [Testing Your Provider](https://github.com/AnalogJ/lexicon/blob/master/CONTRIBUTING.md#testing-your-provider)
- [Test Recordings](https://github.com/AnalogJ/lexicon/blob/master/CONTRIBUTING.md#test-recordings)

If you need any help, I can walk you though any issues you may have.

kapouer:
So if i understand it right, the api token needed to run the tests will be kept known to you and you only ?

AnalogJ:
Hey @kapouer 
Actually no. The API token can be yours, and then the HTTP request and response is recorded by `vcrpy` and stored in https://github.com/AnalogJ/lexicon/tree/master/tests/fixtures/cassettes

When the tests are run by CircleCI or any user, those recordings are used, instead of talking to the API directly, meaning no API token is necessary.

You'll need to ensure that any authentication headers for that provider are removed from the recordings before committing. Heres and example of how I did that with Cloudflare:

https://github.com/AnalogJ/lexicon/blob/master/tests/providers/test_cloudflare.py#L15-L16

Does that all make sense?

kapouer:
Not at all ! I throw the towel now and think `vcrpy` is waaay above the top of my head.

kapouer:
I'll just mimic the other providers without asking questions :)

kapouer:
The `create_record_for_A_with_valid_name_and_content` test (and all tests, actually, but this is the first one) pass `placeholder_priority` as priority (here's what the provider receives: `{'priority': 'placeholder_priority', 'data': '127.0.0.1', 'type': 'A', 'name': 'localhost.edasarl.fr.', 'ttl': 3600}`).
I don't think this is right... It should pass nothing or an integer.

adferrand:
Parameter priority seems to be specific to your provider. Then you should override method _test_engine_overrides in your test suite class, as described in the method documentation from integration_tests, to provide a good default value for priority.

AnalogJ:
By default the `fallbackFn` returns `None`.
However, during testing it will return a `placeholder_` prefixed value to mimic data passed via environmental variables. 

If you want to set a value/default for `priority` then just add the following block in your test file

```
def _test_options(self):
    options = super(OnlineProviderTests, self)._test_options()
    options.update({'provider':'10'}) 
    return options

```

AnalogJ:
If you commit your provider test file I can take a look and ensure it works correctly.

adferrand:
@AnalogJ what is the difference between _test_engine_overrides and _test_options ? Both methods seems to have the same result, provide default value for parameters during tests.

AnalogJ:
Basically its the result of a partly complicated options system. It definitely needs a bit more documentation. 

The relevant runtime source is [here](https://github.com/AnalogJ/lexicon/blob/703b55f27619e68dcc1e207523d977b6f78aff09/lexicon/providers/base.py#L34-L40), test time source is [here](https://github.com/AnalogJ/lexicon/blob/master/tests/providers/integration_tests.py#L316-L328).

Basically, there are 4 types of configuration options
- static defaults - set on every provider, in test and runtime. right now theres only one, `{'ttl': 3600}`.
- engine overrides - used for overriding engine specific options during test time (eg. api endpoint). Empty at runtime. 
- env options - options specified by environmental variables (must be prefixed by `LEXICON_AUTH`). meant for credentials/tokens
- cli options - any options/flags specified via command line parameters (empty at during testing)

kapouer:
it's going better and better... i'm fixing new failures now. Thanks for the help !

kapouer:
I had to work around `test_Provider_when_calling_update_record_should_modify_record` which relies on finding a record by id to return True even if the record was not updated (because it could not find the record by id).
Otherwise tests pass now.

tests/providers/test_online.py
```diff
@@ -0,0 +1,31 @@
...
+	def _filter_headers(self):
+		return ['Authorization']
+
+	def _test_engine_overrides(self):
+		overrides = super(OnlineProviderTests, self)._test_engine_overrides()
+		overrides['fallbackFn'] = (lambda x: 'placeholder_' + x if x != 'priority' else '')
+		return overrides
+
+	@pytest.fixture(autouse=True)
+	def skip_suite(self, request):
```
AnalogJ: Hey, This PR looks just about perfect but it seems you're skipping the extended suite 
Can you remove lines 24 - 27 and ensure that your provider supports record sets?

kapouer: Not that i don't want to, but i'm confused about the meaning of record sets.
In this provider, one can only modify a *disabled* record set, and need to activate it to make it live.
This means, in particular, that the id of a given record will change when the record set it is being part of is made active.
Is this compatible with how lexicon deal with things ?

kapouer: @lexicon maybe i replied that fast that you did not notice :)

AnalogJ: Looking at the documentation, it looks like only zones can be enabled/disabled, not record sets. 

fundamentally the request is that you can have multiple records for same name & type, with differing content.




Type | Name | Content
------------ | ------------- | -------------
A | test1.example.com | 1.2.3.4 
A | test2.example.com | 4.3.2.1
TXT | _acme-challenge.example.com | D-52Wm4V7xoUpGax-F8FrPO45cQRcbRj-XoblaY4uYM
TXT | _acme-challenge.example.com | 6lOgCI0p_LRhtrJMh9aTYAek6hZ64nT75-DkeeQccfA

kapouer: oh right, indeed if record set is only that feature, it's easy to support.

AnalogJ:
Hey @kapouer 
How are the changes to support record sets going? Anything I can help with?

kapouer:
Was busy this week ! I'll give a shot in an hour or two.

kapouer:
Extended tests enabled, rebased, sensitive data removed, squashed in one commit.

AnalogJ:
awesome work. I'll get this merged and released right away.

PR #250, opened 2018-06-21
Allow to specify --content several times

apognu:
This PR allows the ```--content``` argument to be specified more than one. When it is, an array of strings is provided to the provider.

As all current providers expect to receive a string from the CLI, client.py denormalizes one-element arrays that are passed from the CLI into plain strings, for backward compatibility.

### DIFF
--- a/lexicon/client.py
+++ b/lexicon/client.py
@@ -31,7 +31,7 @@ class Client(object):
         self.options.update(cli_options)
 
         # Transform list of content with one element to a plain string to keep backward
-        # compatibility with older provider plugins
+        # compatibility with older provider plugins.
         contents = self.options.get('content', None)
         if contents is not None and len(contents) > 0:
             self.options.update({'content': (contents if len(contents) > 1 else contents[0])})


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/250)

---

apognu:
All tests pass.

I don't know why CircleCI is having difficulties.

AnalogJ:
Hey @apognu 
Thanks for the PR!

If I'm reading this right, the idea is to allow the creation of record sets with one call, rather than requiring the user to call `lexicon create` multiple times. 

I  can definitely see the value, it looks like its designed to be backwards compatible, which is great. We can update the providers over time to support multiple `content` flags.

Let me see if I can get the PR correctly tested on CircleCI, and then merge it if everything works correctly.

AnalogJ:
Hmm, not sure I would quite agree with that.

There are other providers which only support manipulation at the record set level rather than at the record level. 

This limited granularity is usually solved by doing a list operation, finding the current records in the record set, manipulating the set as required, and then submitting the changes to the API.
There's definitely the possibility for a race condition if another tool other than lexicon is updating the records, however it seems to be a fair trade off for not having to know all the records in the record set at runtime. 

If you are creating a Google Cloud provider, I would probably go that route if possible, just to keep all the providers consistent. 

In general I try to ensure that `create` operations are additive, and only `delete` & `update` are destructive. 

Hopefully that all makes sense?

AnalogJ:
weird. I cant quite get CircleCI to retest this PR. 

Can you make another commit to your `dev/multi-contents` branch? CircleCI should automatically test it.

apognu:
I think we agree on the principle, in the provider I am trying to create:
 * ```create``` creates a new record set with (with this patch) one or more ```contents``` (if a record already exists for the same type, it fails with ```409 Conflict```)
 * ```update``` replaces a record set with a whole new content (delete and add in the same transaction)
 * ```delete``` deletes a record set

The only operation Google Cloud DNS does not seem to handle is updating adding a record in an existing record set.

Are you suggesting ```create``` should, if the record set already exists, delete it to recreate it with the new record?

I'll push something on the branch to trigger the build.

apognu:
Is this OK for you?

AnalogJ:
Hey @apognu 

Here's a scenerio table for how I think `create` should work. 

| Scenerio  | Result |
| ------------- | ------------- |
| existing record set, existing record content | do nothing, no exception  |
| existing record set, new record content  | add record to existing record set  |
| record set does not exist, new record content  | create record set with new record  |

Basicall the thought is that record set creation is idempotent, if a record set already exists, always re-use it. If it does not exist, then create a new one. Existing record sets should not throw an error. 

`update` is basically what you wrote. A record set will already exist, so we should just remove the old content and update it with new content if there is only one record in the record set. Unfortunately if there is more than 1 record in a record set `update` doesn't currently work correctly, as we we need the previous record content to determine which record to delete and update.

AnalogJ:
Hey @apognu can you rebase this branch? I've finally fixed the release/packaging automation.

AnalogJ:
this is an outdated PR and conflicts with the current master branch. Closing for now, but I'll reopen if its rebased

PR #261, opened 2018-07-09
Implements Google Cloud DNS provider

adferrand:
This PR implements the Google Cloud DNS provider through the Google API, which is a REST API with OAuth authentication.

Most of the hard work was about to construct a correct OAuth authentication, and to convert the DNS entry model  used by Google to the Lexicon one, as they are completely different.

As always, tests and cassettes are provided for the complete test suite, including record sets.

I think there is some potential discussions about my implementation.

First, I am aware that Google provides a python library to interact with its API. I choosed to not use it for the following reasons:
1) First, the library is about every Google Cloud functionnalities, which is way wider than the sole purpose of Lexicon, which is to interact with Google Cloud DNS: I think it as an overkill to get a library where 95% of the code is not used.
2) Second, as Google states, they advice to use their library unless we develop our own library, which is exactly the case here. Then we can focus only on the requirements to interact with Google DNS, and the way we want. Developpers who want to integrate Lexicon will need only to review Lexicon code, and not all the code in Google library.

Second, as said juste before, we can implement the way we want, and rely on more focused libraries. In fact as OAuth is involved, some advanced cryptographic processings with RSA keys need to be done. I choosed to use the library [cryptography](https://cryptography.io/en/latest/), and to add it not on the provider scope requirement but on the Lexicon global scope requirement. Here are the reasons:
1) About the library choice, since pycrypt development has been dropped, the two main libraries are the fork of pycrypto (pycryptodome) and cryptography. The latter has more than 2500 stars and seems to be widely used and extremely well constructed in terms of security. And so is my choice.
2) About adding it in the global requirements, I think that for every current API and future API implementations which will need strong cryptographic processing (like with OAuth), we can make an opiniated implementation to make developers use this library and rely on its quality.

Finally about the provider auth parameters. There is only one --auth-service-account-info, which embed several things. It is the service account info file generated by Google when we associate the service account to use one the API with a private key. As this file is always provided by Google, it seems to me that it is the natural input to use. But the Lexicon --auth-service-account-info parameter can accept the file path or its content base64 encoded, to make it more portable with Docker containers for instance.

Available for any discussion !

Regards,
Adrien Ferrand


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/261)

---

AnalogJ:
Wow, this is great work 🎉!

> First, I am aware that Google provides a python library to interact with its API. I choosed to not use it for the following reasons ...

Your reasoning is sound. In general my recommendation to re-use existing first-party libraries is primarily because I wanted to easily handle provider API drift. For an API (and company) as large and mature as Google DNS, I doubt there will be much, so we should be safe there. And your point about unnecessary size of their dependency is totally understood.

> Second, as said juste before, we can implement the way we want, and rely on more focused libraries. In fact as OAuth is involved, some advanced cryptographic processings with RSA keys need to be done. I choosed to use the library cryptography, and to add it not on the provider scope requirement but on the Lexicon global scope requirement. Here are the reasons ...

I'm kinda hesitant to include new global libraries because `certbot` has lexicon as a dependency and we've caused breakages to their build pipeline in the past by adding new deps. Obviously I could fix this by just emailing them, but it's still something to consider in general.

---

I was going to mention something again about handing API drift better if we just re-used provider libraries, but I had to remind myself that most of our providers just use `requests` to manually make calls to the API's anyways. I think including/recommending a general purpose OAuth library is probably the right move. 

If you're happy with the PR as is, I can merge it after I ping the @certbot dev team about the new dep.

adferrand:
So, I read the response from certbot devs, seems that it is settled ^^

AnalogJ:
yep, I'm going to merge this and release a new version (minor)

PR #264, opened 2018-07-12
Generate a real output for Lexicon executed in command line

### LINKED ISSUE #251: [CLI] Pretty output for list method
Is there any plans to have pretty outputs (table or at least formatted) for the ```list``` operation on the CLI?

Right now, the CLI assumes a verbosity of DEBUG level, and outputs the Python representation of the result (managed by the provider). If --log_level=ERROR is used, no output is generated, which defeats the CLI usage, in my opinion.

Is this behavior expected? Would you be open to a PR for that?

### LINKED ISSUE #253: Return value for list?
I very well might be looking at this wrong, but it appears that out of the box the "executable" that gets installed for linux can't actually get the list results as a return variable, only as a stdout debug message?

adferrand:
Following issues #251 and #253, I submit this PR to add a real output when we execute Lexicon from command line.

This targets particularly the list command, since it is the main case where we want to get something from the DNS zone, obviously, and since other commands to not provide usable output for now. 

Currently, results obtained during a list are displayed in the form of a JSON through the logging system, and it is displayed to the user because logging is set to debug in command line. However, it carries a lot of debugging information which are not relevant to the main purpose of a list action: get records from the zone.

So first this PR lowers the log level to ERROR in command line, as it is the only level where the command should fail, and so output cannot display a valid response.

Second, the list command display a table if all records found for the given parameters, with a header and a header separator. This is for the human readable response. For the other commands, nothing is returned and we rely on the non 0 return code, which is quite compliant with most gnu utilitaires.

Finally, some additional command line parameters are provided to customize list output in a quite usual way for a shell command. Parameter --quiet removes all the output. Parameter --json returns the json string instead of formatted table, to be used for example through a pipe with jq utility. Parameter --no-headers returns the table without the header and separator, to be used for example with cut utility. 

---

Edited by @AnalogJ:
closes #251 
closes #253


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/264)

---

lexicon/__main__.py
```diff
@@ -60,17 +64,64 @@ def MainParser():
...
     logging.basicConfig(stream=sys.stdout, level=log_level, format='%(message)s')
 
     logger.debug('Arguments: %s', parsed_args)
     client = Client(parsed_args.__dict__)
-    client.execute()
-
+    
+    output = client.execute()
+
+    if not parsed_args.__dict__['quiet']:
```
oldium: Just curious - why not `parsed_args.quiet`?

adferrand: Juste to be consistent with how parsed arguments are passed in the existing code (as a dict).

oldium: Yes, consistency - just look few lines above, here the `parsed_args` is used differently:

```
log_level = logging.getLevelName(parsed_args.log_level)
```

adferrand: Right, I was more focused about the way parsed_args is passed to the provider, as `parsed_args.__dict__`

adferrand: And new commit for that.

AnalogJ:
This is great, something that will actually allow lexicon output to be parsed and utilized by a bash pipe-chain. 

Here's my only feedback.

- Make the function unit-testable (separate print from table generation). I'd like to see a couple of unit tests. It might be lip-service for now, but I'd like to have a starting place to validate changes if we need to refactor the code later. 
- how do you feel about a `--output={json/table/table-no-header}` flag? `--json` and `--no-header` are basically contradictory and can't really be used together anyways.

adferrand:
Option quiet is for the future. Even it does not make much sense for a list operation, it could for create/modify/delete, where we could add some input, like the ID of concerned records.

Yep I also agree that output would be better than several options, and for that matter, 'quiet' could be a valid option for --output.

AnalogJ:
Thanks 🎉!

PR #265, opened 2018-07-12
Subreg.cz: Use Zeep instead of PySimpleSOAP library.

oldium:
PySimpleSOAP ignores certificate errors, so is basically unsecure. Zeep does server certificate verification during SSL setup, so use it instead.

The Zeep library in addition to PySimpleSOAP downloads the WSDL SOAP API description from the URL during instantiation. Alternative would be to supply our own pre-downloaded WSDL and use the Zeep cache feature to simulate already-downloaded file. But for now, keep it simple and let the WSDL be downloaded.

### DIFF
--- a/lexicon/providers/subreg.py
+++ b/lexicon/providers/subreg.py
@@ -10,7 +10,7 @@ import collections
 from .base import Provider as BaseProvider
 
 try:
-    import pysimplesoap # Optional dependency
+    import zeep # Optional dependency
 except:
     pass
 
@@ -26,9 +26,8 @@ class Provider(BaseProvider):
         self.domain_id = None
         self.ssid = None
 
-        self.api = pysimplesoap.client.SoapClient(
-            location="https://subreg.cz/soap/cmd.php?soap_format=1",
-            namespace="http://subreg.cz/types")
+        client = zeep.Client("https://subreg.cz/wsdl")
+        self.api = client.service
 
     # Authenticate against provider,
     # Make any requests required to get the domain's id for this provider, so it can be used in subsequent calls.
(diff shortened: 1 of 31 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/265)

---

AnalogJ:
TBH, I don't have much feedback to give here:

- I'm not familiar with either of those libraries
- we're already using an external dependency for the provider
- I don't personally use Subreg
- You're the original author for the subreg provider (thanks for that btw 👍)

I'll leave this PR open for another couple of days incase anyone else using Subreg has any feedback. If not, I'll merge it (please ping me if I forget)

**Just as a reminder to myself, I'll probably release this as a minor version bump (`v1.X.1`) rather than a patch, since its changing the functionality & dependency of an existing provider.**

oldium:
Regarding the library change - it is really needed. The PySimpleSOAP allows man-in-the-middle attack, because it does not do SSL certificate verification. I found it by testing my Subreg.cz API simulator [subregsim](/oldium/subregsim) over HTTPS - PySimpleSOAP happily connected, it did not care about self-signed certificate. Without patching PySimpleSOAP it is impossible to make it secure. According to (no) responses on PySimpleSOAP github, PySimpleSOAP project is unmaintained. So there are two reasons to move to Zeep - one is that Zeep is working fine with SSL, second one is that it is a maintained project.

I will use the lexicon library to issue Lets Encrypt certificates, so feel free to ask for improvements when lexicon API changes :-)

oldium:
Great, thanks.

Issue #171, opened 2018-02-11
[gandi provider] Add support for new Gandi API (dns.api.gandi.net)


Ezwen:
The new Gandi website comes with a new API available at https://dns.api.gandi.net

New Gandi accounts can only use this API, and not the previous one. Therefore it would be great  to add support for this new Gandi API to lexicon for such users :)

thanks for your work on Lexicon!

AnalogJ:
Hey @Gwend4l 

Thanks for your interest. It looks like @sumpfralle and @adferrand are already working on that in #162 

Once their work is complete, you should be able to select the Live/REST api using a command line flag.

Gwend4l:
Indeed, I had not noticed, thanks for pointing that out! Looking forward to the merge :)

adferrand:
I think that issue can be closed, as we implemented the gandi provider on top of the rest api in #247.

mt7479:
@adferrand Is it possible to specify the protocol with an environment variable ? I tried the following with no luck.

```
export PROVIDER="gandi"
export LEXICON_GANDI_TOKEN="123"
export LEXICON_GANDI_PROTOCOL="rest"

export LEXICON_GANDI_AUTH_TOKEN="123"
export LEXICON_GANDI_API_PROTOCOL="rest"
```

I patched my dehydrated hook and included the necessary command line parameters but environment variables would the nice though.

adferrand:
In fact only options like --auth-* can be specified with environment variables. But protocol is specified with --api-protocol. I named it like this because this option is related to the provider itself, not to the authentication.

@AnalogJ, I think it would be useful to specify any option by environment variables.

mt7479:
Ah, now i understand. It would indeed be nice have this at least for this specific provider, not sure if any other would benefit as well.

nh2:
> I patched my dehydrated hook and included the necessary command line parameters

@mt7479 Can you share your patches? I'm facing the problem

```
Exception: Failed to authenticate: '<Fault 510150: 'Error on object : OBJECT_ACCOUNT (CAUSE_NORIGHT) [Invalid API key]'>'
```

and suspect that it'll go away by using the REST API.

mt7479:
@nh2 Nice hack! 
I wasn't as smart and simply hard-coded all the needed options in the hook file. Including removing the $PROVIDER Variable :-D

adferrand:
PR is on the go to avoid the hacks ^^

adferrand:
[...]

PR #274, opened 2018-08-11
Some evolutions for Gandi provider

adferrand:
Following #171: I replaced the option `--api-protocol` by `--auth-api-protocol`. It allows this option, which drives which Gandi API is used, to be defined by an environment variable (here `LEXICON_GANDI_API_PROTOCOL` because only options formated as `--auth-something` can be handled with environment variables. For retro-compatibility, `--api-protocol` is still read but is invisible, and a warning is displayed to the user, indicating that `--auth-api-protocol` should be used for now on.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/274)

---

AnalogJ:
Hey @adferrand 
I see what you're gettin at. My idea with the `--auth-*` parameters was to semi-secure credentials so that they didn't leak into log files or bash history. 

Rather than rename the `--api-protocol` flag, I think there are 2 more generic solutions:

1. Allow all flags to be passed via environmental variables.
2. Modify the certbot and dehydrated hook scripts to take in a `LEXICON_OPTIONS` environmental variable that can be used by users to pass in arbitrary custom parameters. 

eg. 
```
# modified from: https://github.com/AnalogJ/lexicon/blob/master/examples/certbot.default.sh#L57-L58
lexicon "${PROVIDER}" "${PROVIDER_CREDENTIALS[@]}" \
    create "${CERTBOT_DOMAIN}" TXT --name "_acme-challenge.${CERTBOT_DOMAIN}" --content "${CERTBOT_VALIDATION}" ${LEXICON_OPTIONS}
```

What you think?

adferrand:
Yes, I opened this PR specifically for Gandi, as several users could not enable the REST protocol on Gandi provider only with environment variables.

But this is a more general issue about design around environment variable.

For you solution 1,it is the most obvious answer, but there will be difficulties in understanding which env var name to use for what. For instance, `--api-protocol` would resolve to `LEXICON_GANDI_API_PROTOCOL` and `--auth-token` to `LEXICON_GANDI_TOKEN`. And about options not related to a provider, like `--delegate`?

About the option 2, it is quite explicit, but it is not an implementation at the lexicon level, each developer which uses lexicon will need to do something.

I would like to give you a third solution: `LEXICON_OPTIONS` and `LEXICON_[PROVIDER]_OPTIONS`. For the first one, if specified, options are always added to lexicon, and we can set generic options like so `LEXICON_OPTIONS='--delegate=example.com --whatever=test'`. For the second one, options are added to lexicon when the given provider is used, and we can set provider specific options like so `LEXICON_GANDI_OPTIONS='--api-protocol=rest --whatever-specific=test'`.

adferrand:
Hello Jason. What to you think of my idea?

bmw:
@AnalogJ, I haven't looked through the other parts of this PR, but landing the change to the way the OVH plugin makes its requests would helpful for Certbot.

adferrand:
So as I said to @bmw, I will split my PR in two. 

One for the Gandi stuff, which remains here as it involves further discussions. One for the OVH stuff in a new PR, as it is just a bug correction easily mergeable.

AnalogJ:
Hey @bmw @adferrand 

I've finally gotten the capsulecd release automation working correctly, and I'm able to correctly release new versions of Lexicon again. 

Regarding this PR, yeah I'd prefer to have this split into 2 separate PR's if possible. 

@adferrand 
I've been thinking a bit about env variable support, and I think that it might be a bit difficult to add something generic like `LEXICON_GANDI_OPTIONS` as the CLI flag parsing is done pretty early on by the `ArgumentParser` and I'm not sure we can dynamically inject flags into it. 

I think we might be able to modify the [options parser class](https://github.com/AnalogJ/lexicon/blob/master/lexicon/common/options_handler.py#L12-L17) to parse all compatible arguments, in a backwards compatible way.

1. `LEXICON_*` prefixed env variables are processed, split and lowercased. 
2. `LEXICON_{PROVIDER}_*` prefixed env vars are processed, split and lowercased. 
3. `LEXICON_{PROVIDER}_*` prefixed env vars are processed, split, lowercased and prepended with `auth_`

The `get` command would be modified such that `auth_` prefixed keys print out a STDERR message specifying that that key is deprecated. 

Having said all that, I'm getting a bit frustrated with the limitations of this options parsing system. I actually wanted something like https://github.com/indexzero/nconf but I couldn't find anything that flexible/hierarchical for python. 

I'd be open to a proposal for completely scrapping it and replacing it. At one point I had hoped I could use something as simple as  `parser.add_argument('--auth_token', default=os.environ.get('LEXICON_GANDI_AUTH_TOKEN', None))`

adferrand:
Hello @AnalogJ !

It seems that this story of smart injection of parameters bother you for quite a while ^^

So there is a short term issue and a long term approach to deal with. Let's start with the short-term.

If I understand correctly, I agree with your proposition. Let's consider parameters specified by `--api-protocol` and `--auth-token` for the gandi provider. Internally they are resolved through the `config` object on the keys `api_protocol` and `auth_token`. With what you propose, theses parameters can be specified by the same kind of environment variable pattern, namely `LEXICON_GANDI_API_PROTOCOL` and `LEXICON_GANDI_TOKEN`, because the first one is resolved directly and the second one is resolved by its backward resolution using the prefix `auth_`. And an hypotetical `--token` would also be resolved by `LEXICON_GANDI_TOKEN`.

The thing which is not clear for me is where the deprecation occurs. Is it because:
* we use internally the key `auth_token` ?
* we resolve its value using `LEXICON_GANDI_TOKEN` throught the backward mechanism instead of directly with `LEXICON_GANDI_AUTH_TOKEN` ?
* or because we use `--auth-token` instead of `--token` in the first place ?

I think is option 2, but I would like your confirm before starting to code.

adferrand:
Now for the long term. Yes, an automated and transparent resolution would be great.

I think it is not to complex to make a custom class with three sources of values:
1) Command line argument
2) Environment variable
3) Configuration file

For your study, maybe you will be interested by the specification I have done to rebuild one of my docker (the one with certbot + lexicon), on which I made some thinking about configuration injection: [Project V3 specifications](https://github.com/adferrand/docker-letsencrypt-dns/wiki/Project-V3-specifications).

AnalogJ:
In addition to handing CLI, Env and Files, we would need to support fallbacks/default values as well as the ability to add multiple files, in any order. 

You're right, it shouldn't be too complex, was just something I kept putting off. 

Regarding your V3 specifications, I like the idea of a YAML config file. I use them in other applications I've written.

My only concern is your `profile` concept. While it's definitely more flexible on the letsencrypt implementation side, its not currently supported by lexicon and would require a significant backwards-incompatible change to the env variable names to support it. 

Something like this makes more sense to me (for lexicon)

```
lexicon:
    cloudflare:
      auth_key: "key"
      auth_secret: "secret"
    route53:
      auth_key: "key"
      auth_secret: "secret"
```

or even something flatter that could be generated by bash automation a bit easier (almost like an `.ini` or `.properties` file)

```
lexicon_cloudflare_auth_key: "key"
lexicon_cloudflare_auth_secret: "secret"
...
```

adferrand:
So to sum up, by decreasing priority:
1) Command line argument
2) Environment variable `LEXICON_*` (split + lower)
3) Environment variable `LEXICON_{PROVIDER}_*` (split + lower)
4) Environment variable `LEXICON_{PROVIDER}_*` (split + lower + auth prefix + logger warning)
5) Configuration file (eg. `configB.yml` <- anti-alphabetical order)
6) Configuration file (eg. `configA.yml` <- anti-alphabetical order)
7) Fallback values

About the configuration file, yes, what I made for my Docker is not suitable for Lexicon. I designed it from the point of view of Certbot users, that could manage various providers configuration. No matter what is implemented for Lexicon, I will create a configuration layer in my Docker for its specific needs.

So for Lexicon, it was more about the general idea to define in one place, using a yml file, the parameters for a given provider. In that regard, I think we agree on the yml design. Concerning yml vs flat ini, I still prefer yml. Because:
* as with Python, structural indentation is a gift from Gods :)
* yml is structured, so future evolutions we do not know now will be implemented easier
* even if structured, it is not verbose, contrary to xml or even json, so readability remains good (only two lines more per provider)

If you are good with this specification, I know what to do then ^^

AnalogJ:
hmm, for the priority list I would recommend a couple small changes:

1. Command line argument
2. Environment variable LEXICON_{PROVIDER}_* (split + lower)
3. Environment variable LEXICON_{PROVIDER}_* (split + lower + auth prefix + logger warning)
4. Environment variable LEXICON_* (split + lower)
5. Customizable Configuration file (default. lexicon_{provider}.yml )
6. Customizable Configuration file (default. lexicon.yml)
7. Fallback values

And I would make it configurable. Hard-coded configuration loading has been a huge pain in the ass for me on a couple of different apps. For testing it should be possible to overridate the configuration object so that the following can be passed into the provider under test:

1. Customizable Configuration file (default. lexicon_{provider}.yml )
2. Customizable Configuration file (default. lexicon.yml)
3. Fallback values

Other than that, yeah I agree with your assessment, yaml would definitely be preferrable to ini/properties file style.

adferrand:
Roger that. There is a default resolution path, and we can pass another one anytime we want.

adferrand:
Closed in favor of #302

PR #291, opened 2018-09-21
Tests optimize

adferrand:
Two improvements for the tests:
* Godaddy provider handling on potential 409 errors for a request, by retrying several time the request, because 409 for Godaddy means that DNS zone was under modification. During unit tests, cassettes will replay every 409 response recorded during the live test, increasing drastically the test time execution. But only one 409 is sufficient to test the retry mechanism. I cleaned all Godaddy cassettes to keep only one 409 response for each test, making the Godaddy tests to be executed in an equivalent time to the other providers.
* Parallelize all the things! Using pytest-xdist, tox process launch py.test with `--numprocesses=auto`, so tests are distributed on sub-workers whose number equals the machine CPU one. Option `--dist=loadfile` is added, to ensure that all tests of a given provider are executed sequentially on the same worker, as some workers to not use thread-safe processes (it is the case of transip for instance). With parallel test execution, overall build time drop from 5 min to 2 min 30.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/291)

---

PR #296, opened 2018-10-05
Implementing the auto provider

adferrand:
_One provider to brings them all, and in the `dig` bind them._

Today I would like to propose a special provider: `auto`. As one can assume from its name, this provider does not add support to any new DNS provider. It allows to automatically select the right provider from the given domain name, and to execute the required operation with it.

Basically, one can execute `lexicon auto create mydomain.com TXT --name foo --content bar [credentials_parameters]`, and expects, if the DNS provider is supported by Lexicon, to have the TXT entry be created in the DNS zone without specifying the actual provider.

Credentials for each provider potentially selected by `auto` need to be specified. The `auto` provider support every command line parameter accepted by the other providers. However they need to be prefixed by the provider name concerned, in the form of `--[provider_name]-[parameter_name]`, to avoid name collision of parameters between providers (like the really common `--auth-token`).

So for instance one can pass `--cloudflare-auth-token` to ensure that the cloudflare provider, if used, will have its auth_token. See `lexicon auto --help` for a list of all supported parameters.

Alternatively, and similarly to the other providers, any environment variable in the form of `LEXICON_[PROVIDER_NAME]_[AUTH_PARAMETER_NAME]` will be correctly passed to the relevant provider if `auto` chooses it for a given domain. The syntax is exactly the same as before.

To my mind, the `auto` provider is not really for direct use of the lexicon command line. Because if you execute one particular command, it would be more efficient to use directly the current provider of your DNS zone. In this case, you do not need some kind of automation.

What `auto` is really designed for -and I hope it is where it will really shine- is precisely everywhere automation is required: managing DNS zone portfolio, integrating tools using Lexicon as a library, and of course processes like maintaining wildcard certificates through Certbot for instance ^^. In theses cases, providers potentially used and their authentication assets are known. So with asserts can be passed to Lexicon in a centralized manner, and Lexicon will be able to provide a unified interface other several DNS providers, for large lists of domains, without requiring the user to care about the actual implementation.

Under the wood, how `auto` select the good provider ? One could think about a plain old `WHOIS` request, that gives the registar of a domain. like hundred of sites on the Web do. But the registar is NOT the DNS zone manager (frequently, but not always). Because a registar declares only, for a domain, the nameservers that serves the DNS zone: this is the DNS provider API of theses nameservers that Lexicon needs to target. For instance, DigitalOcean is not a registar, but can manage a zone after a registar declared DigitalOcean nameservers as the nameservers of this zone.

So `auto` finds the nameservers declared for the given domain. It does so by leveraging the call to the `nslookup` binary, available on any distribution (including Windows). Last think is to map the nameservers to the providers available in Lexicon. One could think of a list of all nameservers available for each provider: accurate, but harder to maintain. Nameservers do not change each day, but they change.

What does not change very frequently however, it is the domain of the nameserver FQDN (`ovh.net` in the nameserver `ns1.ovh.net` => LEXICON-CEPTION). So a list of nameserver domains (or regex patterns, thanks AWS) must be exposed by each provider. I did it for all providers I could do. Most of them actually, only 4 of them do not have domains I could find easily. Maybe a couple of them could have errors that will need to be fixed in further patches.

Finally, after domain comparison between found nameservers and declared nameservers for each provider, relevant provider is found, marked as delegate, and any business logic is delegated to it.

Final word about the code architecture: `auto` provider is really a provider. It exposes the exact API that Lexicon expect from a Provider module, without modifying the core logic. I preferred that for maintainability purpose: not everyone will need this functionality, so we should avoid to plug it in the core process. Apart that, `auto` provider makes some complex introspection of the Lexicon runtime, but it is only for its own purpose and will not interfere with other providers.

So, as always, ready to discuss any further improvement on this PR.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/296)

---

AnalogJ:
Haha nice. I'm kinda swamped with some other work right now, but I'll try to get to this ASAP.

lexicon/__main__.py
```diff
@@ -36,13 +38,10 @@ def BaseProviderParser():
...
 def MainParser():
-    current_filepath = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'providers')
-    providers = [os.path.splitext(f)[0] for f in os.listdir(current_filepath) if os.path.isfile(os.path.join(current_filepath, f))]
-    providers = list(set(providers))
-    providers.remove('base')
-    providers.remove('__init__')
-    providers = [x for x in providers if not x.startswith('.')]
-
+    providers = []
+    for _, modname, _ in pkgutil.iter_modules(providers_package.__path__):
```
AnalogJ: this is neat, I haven't seen this before. I'm assuming its supported by python 2.x and 3.x?

lexicon/__main__.py
```diff
@@ -8,10 +8,12 @@
 import os
 import sys
 import json
+import pkgutil
 
 import pkg_resources
 
-from .client import Client
+from lexicon.client import Client
+from lexicon import providers as providers_package
```
AnalogJ: this is where my python ignorance starts to show. Will this still work when using lexicon as a library vs an executable?

lexicon/__main__.py
```diff
@@ -36,13 +38,10 @@ def BaseProviderParser():
...
-    providers = [os.path.splitext(f)[0] for f in os.listdir(current_filepath) if os.path.isfile(os.path.join(current_filepath, f))]
-    providers = list(set(providers))
-    providers.remove('base')
-    providers.remove('__init__')
-    providers = [x for x in providers if not x.startswith('.')]
-
+    providers = []
+    for _, modname, _ in pkgutil.iter_modules(providers_package.__path__):
+        if modname != 'base':
+            providers.append(modname)
```
AnalogJ: do we have to filter out the `__init__` file like we do in the old code?

lexicon/providers/auto.py
```diff
@@ -0,0 +1,173 @@
...
+from builtins import object
+
+import six
+import tldextract
+import sys
+
+from lexicon import providers
+from lexicon.common.options_handler import env_auth_options
+
+LOGGER = logging.getLogger(__name__)
```
AnalogJ: just a nit. can we change this to `logger` like we use in the other classes?

lexicon/providers/auto.py
```diff
@@ -0,0 +1,173 @@
...
+    nameservers declared for the domain, using declared nameservers domain in each Lexicon DNS
+    provider.
+    Any call upon instantiation will be delegated to the resolved Provider.
+    Any command line parameter or environment variable passed to Lexicon will be transfered to
+    the resolved provider if it respect the naming convention: --[provider]-[parameter_name] for a
+    command line parameter, or LEXICON_[PROVIDER]_PARAMETER_NAME for a environment variable.
+    """
+    def __init__(self, options, engine_overrides=None):
+        self.domain = options.get('domain')
+        self.delegate = None
```
AnalogJ: another nit. we already use the term/parameter `delegate` elsewhere in relation to delegated domains. It might be better to use a different variable name here. 

possibly something like: `handler`, `primary`, `operator`, `proxy`? not a big deal though.

lexicon/providers/auto.py
```diff
@@ -0,0 +1,173 @@
...
+                new_options[re.sub('^{0}'.format(target_prefix), '', key)] = value
+            if not key.startswith('auto_'):
+                new_options[key] = value
+
+        new_options['provider_name'] = provider_name
+
+        self.delegate = provider_module.Provider(new_options, self.engine_overrides)
+        self.delegate.authenticate()
+
+    def __getattr__(self, attr_name):
```
AnalogJ: this is smart, good work 👍

lexicon/providers/cloudflare.py
```diff
@@ -18,6 +18,7 @@ def ProviderParser(subparser):
 class Provider(BaseProvider):
 
     def __init__(self, options, engine_overrides=None):
+        print(options)
```
AnalogJ: extra print statement. should be removed.

requirements.txt
```diff
@@ -6,7 +6,5 @@
 #
 ###############################################################################
 
-requests[security]
-tldextract
-future
-cryptography
+--process-dependency-links
+.
```
AnalogJ: nice. I've been meaning to do this for a while. 

If I remember correctly though, we'll need to add `requests[security]` as a dependency for lexicon in setup.py.

tox.ini
```diff
@@ -21,6 +21,7 @@ deps =
 passenv = CIRCLE_BRANCH
 commands =
     lexicon --version
+    tldextract --update
```
AnalogJ: where does this binary come from?

AnalogJ:
Here's another thing I think we might need to consider:

Split horizon DNS. Its definitely kind of niche, but definitely used at larger companies. ie. `*.corp.company.com` served by internal DNS while `*.company.com` resolved by public dns. 

What happens if they select auto, but try to update a record for `internalsite.corp.company.com`

lexicon/__main__.py
adferrand: In fact, this is even the best way to ensure that it will work outside the development area (so either as standalone executable or imported as library). In general, people say to use absolute import and relative import for corner situations (like if you need to dig into 10 levels of packages to get a module, but in this case, there is something to think about the library design ^^).

Because absolute import are not tied to the actual filesystem architecture of a given library.

The only thing to care about correctly is during development. Because in this case, the starting directory may influence whether a package is accessible or not. But you completely mitigate that if you set up your dev environnement "by the book", ie by invoking first `pip install -e lexicon_source_path`. Then lexicon package is correctly exposed, independently of its actual location, and you match your dev env with the deployed env. 

This what is advised in your CONTRIBUTING.md, so we are covered.

lexicon/__main__.py
adferrand: Yes, it is working on both. I developed using both py27 on Debian and py37 on Windows. And the CI is testing both versions anyway.

lexicon/__main__.py
adferrand: In fact your question is closely related to the other one about absolute imports and the fact that absolute imports are not tied to an actual filesystem architecture. `pkgutil.iter_modules` does not care about the Python files, it cares about what is resolved as packages and modules in the current Python runtime.

Namely, it is: 'iterate over all the modules available in the given package'. In that extend, `__init__` does not exists. It is just a file containing the logic to construct the relevant package, but it is not itself a module. So it does not appear in the iteration.

What are the other advantages ? Well, imagine that something (certainly `__init__.py`) construct and dynamically add a module in a package. If you rely on the filesystem, you will not see the dynamic module, because there is not file for it. But if you call `pkgutil.iter_modules`, you will see it.

lexicon/providers/auto.py
adferrand: Of course. Next commit.

lexicon/providers/auto.py
adferrand: I will tag `proxy_provider` to be explicit about the nature of the proxy. Next commit.

lexicon/providers/auto.py
adferrand: Thanks ! For the record, you could ask why it did not placed the provider proxy resolution logic in the constructor. It is because of Certbot. Last time I pushed some advanced logic in the constructor, Certbot tests could fail because their mock are enabled only on the provider methods, not its constructor.


Quite related in the Lexicon tests. I put some effort to not rely on an actual execution of the `nslookup` binary by mocking the relevant method call, and make the tests more portable. The only actual execution of `nslookup` is done on an explicit test designed for that, and here again I skip it automatically for known situations where it cannot work correctly (no network on the test system for instance).

lexicon/providers/cloudflare.py
adferrand: Right. Next commit.

adferrand:
Good point.

Correct behavior all depends on which nameserver is returned when you nslookup something like `*.corp.company.com`. If I understand correctly, when split-horizon DNS is enabled, any call to the primary zone `company.com.` concerning a record of `corp.company.com` will be transparently delegated to the zone holding `corp.company.com`: so it is the nameservers of `corp.company.com` that will be returned, and `auto` provider will be able to understand that it cannot manage this zone.

For that, it means that `auto` needs to ask for `corp.company.com`, not `company.com`. Next question so is what is the `domain` value in the `options` dict provided to the provider constructor. This is handled by the `Client` constructor.

I see from the code that `Client` takes into account the `--delegated` parameter provided by the user, and appends it to the `domain` calculated by `tldextract` (which is `company.com`).

So, if the user passes correctly `--delegated corp`, the effective domain passed to the provider will be `corp.company.com`. That is required anyway for any provider on a delegated domain, so everything is good to me.

requirements.txt
adferrand: Done. Committed.

adferrand:
Hello @AnalogJ, I stay available to talk and improve the provider ^^

AnalogJ:
Hey @adferrand sorry about the delay. I'm traveling right now, but I'll try to take a look at this again this week.

AnalogJ:
Took another look at this, and everything looks good. I'll release a new version.

adferrand:
All right! Once this is merged, I will need some extra work on my other PR about extensible config to make auto provider take advantage of it.

AnalogJ:
Yeah, and we'll also need to make sure that all subsequent PR's follow the same pattern. 

I wonder if it would be possible to add unit tests to ensure that every provider has a entry in the CodeOwners file and that they all use fully qualified package names rather than relative names, ie. `lexicon.providers.base`

AnalogJ:
hmm. capsulecd is complaining while running the test command

``` >> 2018/10/28 20:56:39 stdout py27 runtests: commands[1] | tldextract --update
 >> 2018/10/28 20:56:39 stdout py27 runtests: commands[2] | py.test tests --cov=lexicon --numprocesses=auto --dist=loadfile
 >> 2018/10/28 20:56:40 stdout ============================= test session starts ==============================
 >> 2018/10/28 20:56:40 stdout platform linux2 -- Python 2.7.15, pytest-3.8.0, py-1.7.0, pluggy-0.8.0
 >> 2018/10/28 20:56:40 stdout rootdir: /tmp/480000876/lexicon, inifile:
 >> 2018/10/28 20:56:40 stdout plugins: xdist-1.23.0, forked-0.2, cov-2.6.0
 >> 2018/10/28 20:56:40 stdout gw0 I
 >> 2018/10/28 20:56:44 stdout gw0 [1296]
 >> 2018/10/28 20:56:44 stdout
 >> 2018/10/28 20:56:44 stdout scheduling tests via LoadFileScheduling
 >> 2018/10/28 20:56:45 stdout ....................................................ss.................. [  5%]
 >> 2018/10/28 20:56:49 stdout ........ss.........F.................ss......s...................ss..... [ 11%]
 >> 2018/10/28 20:56:51 stdout .....................ss......s...................ss..................... [ 16%]
 >> 2018/10/28 20:56:54 stdout ...s.ss......s...................ss..........................ss......s.. [ 22%]
 >> 2018/10/28 20:56:56 stdout .......ss....ssssss..s...s.........ss....ssssss..s...s.........ss....ss. [ 27%]
 >> 2018/10/28 20:56:58 stdout sss..s...s...................ss..........................ss............. [ 33%]
 >> 2018/10/28 20:57:01 stdout .............ss................ss....ss.sss..s.............ss....ss.sss. [ 38%]
 >> 2018/10/28 20:57:06 stdout .s...s...................ss..........................ss................. [ 44%]
 >> 2018/10/28 20:57:16 stdout .........ss..........................ss........................s.ss..... [ 50%]
 >> 2018/10/28 20:57:19 stdout ...................s.ss............s.............ss......s.............. [ 55%]
 >> 2018/10/28 20:57:22 stdout .....ss......s.................s.ss........................s.ss......... [ 61%]
 >> 2018/10/28 20:57:25 stdout .................ss......s.................s.ss......s.................. [ 66%]
 >> 2018/10/28 20:57:29 stdout .ss.....................s....ss..........................ss............. [ 72%]
 >> 2018/10/28 20:57:32 stdout ...........s.ss........................s.ss......s.........ss....ss.sss. [ 77%]
 >> 2018/10/28 20:57:35 stdout .s...s...................ss..........................ss.....ss.s.......s [ 83%]
 >> 2018/10/28 20:57:40 stdout s...sss.sss..s.............ss....ss.sss..s.......................ss..... [ 88%]
 >> 2018/10/28 20:57:57 stdout .......s........s....ss................ss....ss.sss..s...s.............. [ 94%]
 >> 2018/10/28 20:57:59 stdout .....ss......s.................s.ss................ss....ss.sss..s...s.. [100%]Coverage.py warning: No data was collected. (no-data-collected)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout =================================== FAILURES ===================================
 >> 2018/10/28 20:58:00 stdout __________________ OvhProviderTests.test_nslookup_resolution ___________________
 >> 2018/10/28 20:58:00 stdout [gw0] linux2 -- Python 2.7.15 /tmp/480000876/lexicon/.tox/py27/bin/python
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout self = <test_auto.OvhProviderTests testMethod=test_nslookup_resolution>
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout     @pytest.mark.skipif(there_is_no_network(), reason='No network, no nslookup call possible.')
 >> 2018/10/28 20:58:00 stdout     @pytest.mark.ignore_nslookup_mock('yes')
 >> 2018/10/28 20:58:00 stdout     def test_nslookup_resolution(self):
 >> 2018/10/28 20:58:00 stdout         """Ensure that nameservers can be resolved through os nslookup call."""
 >> 2018/10/28 20:58:00 stdout >       assert _get_ns_records_domains_for_domain('google.com')
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout tests/providers/test_auto.py:60:
 >> 2018/10/28 20:58:00 stdout _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 >> 2018/10/28 20:58:00 stdout lexicon/providers/auto.py:39: in _get_ns_records_domains_for_domain
 >> 2018/10/28 20:58:00 stdout     tlds = [tldextract.extract(ns_entry) for ns_entry in _get_ns_records_for_domain(domain)]
 >> 2018/10/28 20:58:00 stdout lexicon/providers/auto.py:47: in _get_ns_records_for_domain
 >> 2018/10/28 20:58:00 stdout     stderr=subprocess.STDOUT)
 >> 2018/10/28 20:58:00 stdout /usr/local/lib/python2.7/subprocess.py:216: in check_output
 >> 2018/10/28 20:58:00 stdout     process = Popen(stdout=PIPE, *popenargs, **kwargs)
 >> 2018/10/28 20:58:00 stdout /usr/local/lib/python2.7/subprocess.py:394: in __init__
 >> 2018/10/28 20:58:00 stdout     errread, errwrite)
 >> 2018/10/28 20:58:00 stdout _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout self = <subprocess.Popen object at 0x7f9cdabd6790>
 >> 2018/10/28 20:58:00 stdout args = ['nslookup', '-querytype=NS', 'google.com'], executable = 'nslookup'
 >> 2018/10/28 20:58:00 stdout preexec_fn = None, close_fds = False, cwd = None, env = None
 >> 2018/10/28 20:58:00 stdout universal_newlines = False, startupinfo = None, creationflags = 0, shell = False
 >> 2018/10/28 20:58:00 stdout to_close = set([20]), p2cread = None, p2cwrite = None, c2pread = 20
 >> 2018/10/28 20:58:00 stdout c2pwrite = 22, errread = None, errwrite = 22
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout     def _execute_child(self, args, executable, preexec_fn, close_fds,
 >> 2018/10/28 20:58:00 stdout                        cwd, env, universal_newlines,
 >> 2018/10/28 20:58:00 stdout                        startupinfo, creationflags, shell, to_close,
 >> 2018/10/28 20:58:00 stdout                        p2cread, p2cwrite,
 >> 2018/10/28 20:58:00 stdout                        c2pread, c2pwrite,
 >> 2018/10/28 20:58:00 stdout                        errread, errwrite):
 >> 2018/10/28 20:58:00 stdout         """Execute program (POSIX version)"""
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout         if isinstance(args, types.StringTypes):
 >> 2018/10/28 20:58:00 stdout             args = [args]
 >> 2018/10/28 20:58:00 stdout         else:
 >> 2018/10/28 20:58:00 stdout             args = list(args)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout         if shell:
 >> 2018/10/28 20:58:00 stdout             args = ["/bin/sh", "-c"] + args
 >> 2018/10/28 20:58:00 stdout             if executable:
 >> 2018/10/28 20:58:00 stdout                 args[0] = executable
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout         if executable is None:
 >> 2018/10/28 20:58:00 stdout             executable = args[0]
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout         def _close_in_parent(fd):
 >> 2018/10/28 20:58:00 stdout             os.close(fd)
 >> 2018/10/28 20:58:00 stdout             to_close.remove(fd)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout         # For transferring possible exec failure from child to parent
 >> 2018/10/28 20:58:00 stdout         # The first char specifies the exception type: 0 means
 >> 2018/10/28 20:58:00 stdout         # OSError, 1 means some other error.
 >> 2018/10/28 20:58:00 stdout         errpipe_read, errpipe_write = self.pipe_cloexec()
 >> 2018/10/28 20:58:00 stdout         try:
 >> 2018/10/28 20:58:00 stdout             try:
 >> 2018/10/28 20:58:00 stdout                 with self._disabling_gc_lock:
 >> 2018/10/28 20:58:00 stdout                     gc_was_enabled = gc.isenabled()
 >> 2018/10/28 20:58:00 stdout                     # Disable gc to avoid bug where gc -> file_dealloc ->
 >> 2018/10/28 20:58:00 stdout                     # write to stderr -> hang.
 >> 2018/10/28 20:58:00 stdout                     # https://bugs.python.org/issue1336
 >> 2018/10/28 20:58:00 stdout                     gc.disable()
 >> 2018/10/28 20:58:00 stdout                 try:
 >> 2018/10/28 20:58:00 stdout                     self.pid = os.fork()
 >> 2018/10/28 20:58:00 stdout                 except:
 >> 2018/10/28 20:58:00 stdout                     if gc_was_enabled:
 >> 2018/10/28 20:58:00 stdout                         gc.enable()
 >> 2018/10/28 20:58:00 stdout                     raise
 >> 2018/10/28 20:58:00 stdout                 self._child_created = True
 >> 2018/10/28 20:58:00 stdout                 if self.pid == 0:
 >> 2018/10/28 20:58:00 stdout                     # Child
 >> 2018/10/28 20:58:00 stdout                     try:
 >> 2018/10/28 20:58:00 stdout                         # Close parent's pipe ends
 >> 2018/10/28 20:58:00 stdout                         if p2cwrite is not None:
 >> 2018/10/28 20:58:00 stdout                             os.close(p2cwrite)
 >> 2018/10/28 20:58:00 stdout                         if c2pread is not None:
 >> 2018/10/28 20:58:00 stdout                             os.close(c2pread)
 >> 2018/10/28 20:58:00 stdout                         if errread is not None:
 >> 2018/10/28 20:58:00 stdout                             os.close(errread)
 >> 2018/10/28 20:58:00 stdout                         os.close(errpipe_read)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         # When duping fds, if there arises a situation
 >> 2018/10/28 20:58:00 stdout                         # where one of the fds is either 0, 1 or 2, it
 >> 2018/10/28 20:58:00 stdout                         # is possible that it is overwritten (#12607).
 >> 2018/10/28 20:58:00 stdout                         if c2pwrite == 0:
 >> 2018/10/28 20:58:00 stdout                             c2pwrite = os.dup(c2pwrite)
 >> 2018/10/28 20:58:00 stdout                         if errwrite == 0 or errwrite == 1:
 >> 2018/10/28 20:58:00 stdout                             errwrite = os.dup(errwrite)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         # Dup fds for child
 >> 2018/10/28 20:58:00 stdout                         def _dup2(a, b):
 >> 2018/10/28 20:58:00 stdout                             # dup2() removes the CLOEXEC flag but
 >> 2018/10/28 20:58:00 stdout                             # we must do it ourselves if dup2()
 >> 2018/10/28 20:58:00 stdout                             # would be a no-op (issue #10806).
 >> 2018/10/28 20:58:00 stdout                             if a == b:
 >> 2018/10/28 20:58:00 stdout                                 self._set_cloexec_flag(a, False)
 >> 2018/10/28 20:58:00 stdout                             elif a is not None:
 >> 2018/10/28 20:58:00 stdout                                 os.dup2(a, b)
 >> 2018/10/28 20:58:00 stdout                         _dup2(p2cread, 0)
 >> 2018/10/28 20:58:00 stdout                         _dup2(c2pwrite, 1)
 >> 2018/10/28 20:58:00 stdout                         _dup2(errwrite, 2)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         # Close pipe fds.  Make sure we don't close the
 >> 2018/10/28 20:58:00 stdout                         # same fd more than once, or standard fds.
 >> 2018/10/28 20:58:00 stdout                         closed = { None }
 >> 2018/10/28 20:58:00 stdout                         for fd in [p2cread, c2pwrite, errwrite]:
 >> 2018/10/28 20:58:00 stdout                             if fd not in closed and fd > 2:
 >> 2018/10/28 20:58:00 stdout                                 os.close(fd)
 >> 2018/10/28 20:58:00 stdout                                 closed.add(fd)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         if cwd is not None:
 >> 2018/10/28 20:58:00 stdout                             os.chdir(cwd)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         if preexec_fn:
 >> 2018/10/28 20:58:00 stdout                             preexec_fn()
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         # Close all other fds, if asked for - after
 >> 2018/10/28 20:58:00 stdout                         # preexec_fn(), which may open FDs.
 >> 2018/10/28 20:58:00 stdout                         if close_fds:
 >> 2018/10/28 20:58:00 stdout                             self._close_fds(but=errpipe_write)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                         if env is None:
 >> 2018/10/28 20:58:00 stdout                             os.execvp(executable, args)
 >> 2018/10/28 20:58:00 stdout                         else:
 >> 2018/10/28 20:58:00 stdout                             os.execvpe(executable, args, env)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                     except:
 >> 2018/10/28 20:58:00 stdout                         exc_type, exc_value, tb = sys.exc_info()
 >> 2018/10/28 20:58:00 stdout                         # Save the traceback and attach it to the exception object
 >> 2018/10/28 20:58:00 stdout                         exc_lines = traceback.format_exception(exc_type,
 >> 2018/10/28 20:58:00 stdout                                                                exc_value,
 >> 2018/10/28 20:58:00 stdout                                                                tb)
 >> 2018/10/28 20:58:00 stdout                         exc_value.child_traceback = ''.join(exc_lines)
 >> 2018/10/28 20:58:00 stdout                         os.write(errpipe_write, pickle.dumps(exc_value))
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                     finally:
 >> 2018/10/28 20:58:00 stdout                         # This exitcode won't be reported to applications, so it
 >> 2018/10/28 20:58:00 stdout                         # really doesn't matter what we return.
 >> 2018/10/28 20:58:00 stdout                         os._exit(255)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout                 # Parent
 >> 2018/10/28 20:58:00 stdout                 if gc_was_enabled:
 >> 2018/10/28 20:58:00 stdout                     gc.enable()
 >> 2018/10/28 20:58:00 stdout             finally:
 >> 2018/10/28 20:58:00 stdout                 # be sure the FD is closed no matter what
 >> 2018/10/28 20:58:00 stdout                 os.close(errpipe_write)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout             # Wait for exec to fail or succeed; possibly raising exception
 >> 2018/10/28 20:58:00 stdout             data = _eintr_retry_call(os.read, errpipe_read, 1048576)
 >> 2018/10/28 20:58:00 stdout             pickle_bits = []
 >> 2018/10/28 20:58:00 stdout             while data:
 >> 2018/10/28 20:58:00 stdout                 pickle_bits.append(data)
 >> 2018/10/28 20:58:00 stdout                 data = _eintr_retry_call(os.read, errpipe_read, 1048576)
 >> 2018/10/28 20:58:00 stdout             data = "".join(pickle_bits)
 >> 2018/10/28 20:58:00 stdout         finally:
 >> 2018/10/28 20:58:00 stdout             if p2cread is not None and p2cwrite is not None:
 >> 2018/10/28 20:58:00 stdout                 _close_in_parent(p2cread)
 >> 2018/10/28 20:58:00 stdout             if c2pwrite is not None and c2pread is not None:
 >> 2018/10/28 20:58:00 stdout                 _close_in_parent(c2pwrite)
 >> 2018/10/28 20:58:00 stdout             if errwrite is not None and errread is not None:
 >> 2018/10/28 20:58:00 stdout                 _close_in_parent(errwrite)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout             # be sure the FD is closed no matter what
 >> 2018/10/28 20:58:00 stdout             os.close(errpipe_read)
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout         if data != "":
 >> 2018/10/28 20:58:00 stdout             try:
 >> 2018/10/28 20:58:00 stdout                 _eintr_retry_call(os.waitpid, self.pid, 0)
 >> 2018/10/28 20:58:00 stdout             except OSError as e:
 >> 2018/10/28 20:58:00 stdout                 if e.errno != errno.ECHILD:
 >> 2018/10/28 20:58:00 stdout                     raise
 >> 2018/10/28 20:58:00 stdout             child_exception = pickle.loads(data)
 >> 2018/10/28 20:58:00 stdout >           raise child_exception
 >> 2018/10/28 20:58:00 stdout E           OSError: [Errno 2] No such file or directory
 >> 2018/10/28 20:58:00 stdout
 >> 2018/10/28 20:58:00 stdout /usr/local/lib/python2.7/subprocess.py:1047: OSError
 >> 2018/10/28 20:58:00 stdout```

adferrand:
Ah. It means that `nslookup` is not installed on the system that runs CapsuleCD. I thought it was a quite reasonable dependency to assume to be installed, but maybe not. I mock the call to `nslookup` for every test expect one, to have an actual test on the correct behavior of `nslookup`. 

I see two options to solve that:
* either we do not trigger the test on systems that do not have `nslookup`, and we put an explicit exception on systems that do not have it when `auto` provider is invoked
* or we implement a fallback mechanism to ask remote nslookup sites to do the job. There is one or two available, with strong limits about how many time you can call them in a given period of time.

I would take the first one, as it can be solved elegantly by the user: it is a really basic binary dependency to install on your system, if it is not installed by default. But it is your call. What do you prefer?

In parallel, you can modify the test environment to install `nslookup` and ensure this test is correctly execute, for any option you choose.

adferrand:
About relative imports however, you cannot make pylint consider them as errors, because it is valid python code. Il will complain however if you try to make implicit relative imports, but not for explicit relative imports.

For the former, only a code review can make you deal with this.

Issue #313, opened 2018-11-05
OVH provider does not support proxy


lbillich:
If you set a proxy via environment variables, the OVH provider does not use it.
I think it's because the OVH provider is directly using the method Session.send instead of using Session.request (in the method _request of ovh.py).
The method Session.request support the proxy if it has been set via environment variables.
The method Session.send does not.

AnalogJ:
closed.

PR #314, opened 2018-11-05
OVH provider: add proxy support

lbillich:
Use session.request instead of session.send to support proxy.

No regression in "py.test tests/providers/test_ovh.py".

This is a correction for #313

### DIFF
--- a/lexicon/providers/ovh.py
+++ b/lexicon/providers/ovh.py
@@ -206,9 +206,15 @@ class Provider(BaseProvider):
         ]).encode('utf-8'))
 
         # Sign the request
-        prepared_request.headers['X-Ovh-Signature'] = '$1$' + signature.hexdigest()
+        headers['X-Ovh-Signature'] = '$1$' + signature.hexdigest()
 
-        result = self.session.send(prepared_request)
+        result = self.session.request(
+            method=action,
+            url=target,
+            params=query_params,
+            data=body,
+            headers=headers
+        )
         result.raise_for_status()
 
         return result.json()


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/314)

---

lbillich:
I'm having the proxy problem with certbot when I use the ovh plugin.
Will you PR certbot to integrate your next release?

lbillich:
My PR might not work. I'm in the process of testing it with the real servers and I have an error:

`Error adding TXT record: 400 Client Error: Bad Request for url: https://eu.api.ovh.com/1.0/domain/zone/mydomain.ovh/record?fieldType=TXT&subDomain=_acme-challenge.sub`

lbillich:
I've reworked the correction and now it's working correctly.

adferrand:
I remembered that I had serious problems with OVH to accept the requests, even if they appeared to be well formatted. That is why I used a session for all requests, instead of making individual requests out of an http session, like the other providers do.

Pay a particular attention to the timestamp that is used in the request, it needs to be correctly shifted from the OVH local time. Maybe you dropped the relevant logic accidentally when using session.request instead of request.send.

For that matter, if you succeed in using session.request, maybe you could drop entirely the session, and use bare requests.request instead. A good start for that would be to move the request to the OVH server time that is currently in authenticate() method, inside the _request method, to get correct time each time you make a request.

Obviously you will need to regenerate the cassettes in this cas for integration test replay. See contributing.md to know the environment variable to set and activate live recording against the real API during integration tests.

lbillich:
I'm still using the session.
The error in my correction was only with the signature. I was using the url without its parameters.
My new correction is working correctly, I've tested it with certbot with the real server.

adferrand:
All right ! Would you mind if you could test without the session ?

lbillich:
I've tried it really quick after what you said before but it didn't worked. I didn't check why exactly.

I'm pleased with the current correction and I have created a patch file for ovh.py to apply in my certbot Docker image.

adferrand:
No problem, I understand, if it works, no need to push further.

About deployment now, @AnalogJ, could you make a new minor version before the new Certbot release ? This is in the next 24 hours, I know it is short ... If you can, I will immediately make the relevant PRs on Certbot project to use the corrected version of OVH provider.

lbillich:
You can wait until the next release if you prefer because I already put a patch in place in my Dockerfile.

adferrand:
Thank you ^^ I will try to push it because it's cleaner, but no stress so.

AnalogJ:
LGTM, I'll release a new version momentarily.

PR #325, opened 2018-12-04
add Hetzner provider

rembik:


adferrand:
Do not push to hard because of the lint failure. It is broken my giving a false low grade. I will make a PR tomorrow to fix that.

rembik:
Thx.

rembik:
@adferrand do you need something more to merge this provider to a new release?

adferrand:
@rembik I am not the maintainer of the project, just a regular contributer. It is @AnalogJ that is responsible for that, and he will do soon ^^

For the lint problem, it is fixed, my PR as been merged. You can merge from master to see the corrected lint process be executed, and hopefully becoming green.

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+            LOGGER.error('Hetzner => Unable to authenticate session with %s account \'%s\': '
+                         'Invalid credentials',
+                         self.account, username)
+            raise AssertionError
+        LOGGER.info('Hetzner => Authenticate session with %s account \'%s\'',
+                    self.account, username)
+        return session
+
+    def _exit_session(self):
+        if self._get_provider_option('live_tests') is None:
```
adferrand: Code is aware to be executed in a test. See my comment for this topic below.

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+            if content[0] != '"':
+                content = '"' + content
+            if content[-1] != '"':
+                content += '"'
+        if rdtype in ('CNAME', 'MX', 'NS', 'SRV'):
+            if content[-1] != '.':
+                content = self._fqdn_name(content)
+        return content
+
+    def _raw_content(self, rdtype, content):
```
adferrand: This function is already implemented by the BaseProvider: _clean_TXT_record

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+                        return True
+                retry += 1
+                LOGGER.info('Hetzner => Record is not propagated, %d retries remaining - '
+                            'wait 30s...', (max_retry - retry))
+                time.sleep(30)
+        return False
+
+    # DNS Helpers
+    ###########################################################################
+    def _dns_lookup(self, qname, rdtype, nameservers=None):
```
adferrand: For this matter, in the auto provider, I made a direct call to the nslookup binary. I do not know if it is better to add an optional dependency to dnspython and call it, or to use nslookup.

But we should choose an approach, and be in sync throughout lexicon code.

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+        rrset = self.zone['data'].get_rdataset((self.cname if self.cname
+                                                else self._fqdn_name(name)),
+                                               rdtype=type, create=True)
+        for rdata in rrset:
+            if self._wellformed_content(type, content) == rdata.to_text():
+                LOGGER.info('Hetzner => Record with content \'%s\' already exists', content)
+                self._exit_session()
+                return True
+
+        ttl = (rrset.ttl if rrset.ttl > 0 and rrset.ttl < self._get_lexicon_option('ttl')
```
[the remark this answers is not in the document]
rembik: This means if an recordset with an lower TTL already exists and get one more record, it takes the lower one. The restriction is one TTL on recordset level not on record level.

lexicon/providers/hetzner.py
rembik: I need dnspython for zone manipulating anyway. And with dnspython I know how to use another nameserver as resolver for dns lookups, what - I think - gives me the opportunity to get the propagated state for new records faster.

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+                                                params=query_params, data=data)
+                response.raise_for_status()
+                break
+            except requests.exceptions.ConnectionError:
+                if retry == 9:
+                    raise requests.exceptions.ConnectionError
+                time.sleep(1)
+        return response
+
+    @staticmethod
```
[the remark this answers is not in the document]
rembik: Take it from you 👍

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+        return True
+
+    # Lexicon Helpers
+    ###########################################################################
+    def _request(self, action='GET', url='/', data=None, query_params=None):
+        if data is None:
+            data = {}
+        if query_params is None:
+            query_params = {}
+        for retry in range(10):
```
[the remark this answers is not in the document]
rembik: Hetzner has an really short keep-alive value for sessions. When the connection is broken requests session reconnecting automatically, but during the vcr tests it through an error. Not sure if your approach is working. Could give it a try.

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,604 @@
...
+                                                        self._wellformed_content(type, content))
+                rrset.update(renew_rdataset)
+            synced_change = self._post_zone()
+            if synced_change:
+                self._propagated(type, name, content)
+            self._exit_session()
+            return synced_change
+
+        LOGGER.error('Hetzner => Record lookup has no matches')
+        self._exit_session()
```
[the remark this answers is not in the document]
rembik: Dont really need to close session, but it seems cleaner for me.

lexicon/providers/hetzner.py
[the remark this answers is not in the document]
adferrand: Indeed. A sane assumption in this case. You avoid to modify implicitly ttl of records that are not concerned by the current action.

lexicon/providers/hetzner.py
rembik: I do not know how to change this. At this point I want to avoid that the test script exit the session because all action calls will be done with only one authenticate call and the second time I use the live_tests env var it fasten up the offline test for Hetzner Robot.
There is no given option to manipulate, because only the abnormal way the actions are executed in the test is the reason the script fails. I would need to create one anyway, ar'nt I?

adferrand: Your solution overcome the problem for the tests, but you will still have it outside tests. Because lexicon is used as a library, not only as a cli. And as a library, a developer would expect to create a lexicon client, call the authenticate method, and then execute several actions. So it will fail, because the session will be claused after the first action.

I think it is a design problem. You want to open the session during the authentication phase. You should open the session for each action then close at the end. Like this, each action is independant from each other, and all other problems that are related to the ending session handling that I noted would also vanish. 

The question now is: why do you want to create the session during the authentication phase, and is it a problem to have it open and closed several time if there is several actions (like for instance during a create/update/delete that will need also to call a list at some point).

If it is only a problem of save some requests,I will say this to you: I also tried to do it with ovh, it gave me only problems. Better to have some requests more, less speed, and have a complete isolation between actions.

lexicon/providers/hetzner.py
adferrand: If you need it to manipulate the zone, then it will be the way. However, relying on another dns server than the one set for the host could be a problem. In a corporate protected network, you are not sure that you could touch an external dns server appart the internal one set by the network administrators.

rembik:
Locally on python 3.6 and 2.7 all tests run without an error. Did the pylint fix you told about broke all given cassettes?

lexicon/providers/hetzner.py
rembik: Maybe I should use the local resolver as the initial one...

adferrand:
Hum. Seems unrelated issues. First for lint, it is because of this:
```
lexicon/providers/hetzner.py:434:19: E1101: Instance of 'Name' has no 'labels' member (no-member)
```

adferrand:
> Hum. Seems unrelated issues. First for lint, it is because of this:
> 
> ```
> lexicon/providers/hetzner.py:434:19: E1101: Instance of 'Name' has no 'labels' member (no-member)
> ```

I already had this problem, dnspython use factories that copies dynamically methods and arguments to the constructed object. Pylint cannot see that, you will need to put some a ignore rule on this line.

adferrand:
@rembik, a new version of the module used by localzone has been released by @ags-slc. By pulling the new dependency, integrations tests should work know. You can retry the build.

lexicon/providers/hetzner.py
rembik: Yeaahhh, you nailed it! ^^ I changed it. After that I think there is only one point left. How to fasten up offline test. Now on every create|update|delete call who changes the zone file on Hetzner Robot, there will be a delay of 30s, because in live mode Hetzner need this time to took over the updated zone file.

lexicon/providers/hetzner.py
```diff
@@ -0,0 +1,607 @@
...
+                         self.domain_id,
+                         self.zone['data'].to_text(relativize=True).decode('UTF-8'))
+            return False
+
+        LOGGER.info('Hetzner => Update data for domain ID %s\n\n%s',
+                    self.domain_id,
+                    self.zone['data'].to_text(relativize=True).decode('UTF-8'))
+        if self.account == 'robot':
+            LOGGER.info('Hetzner => Wait 30s until Hetzner Robot has took over data...')
+            time.sleep(30)
```
rembik: How to fasten up offline test? Now on every create|update|delete call who changes the zone file on Hetzner Robot, there will be a delay of 30s, because in live mode Hetzner need this time to took over the updated zone file.

lexicon/providers/hetzner.py
adferrand: I add a similar problem with godaddy, because the API does not propagate immediately the changes. The zone was in a pending state, and you should poll until the API respond a correct status. So during the replay, requests retry capability was also triggered by the recorded responses that would return multiple time the pending status before responding a valid status.

Well, I solved this manually. After recording, and before merging, I just remove almost all pending status response, except on or two for a few tests, to ensure the retry logic is still triggered.

lexicon/providers/hetzner.py
adferrand: See my response above.

adferrand: It implies to poll until the changes are propagated, if a part of the API allow you to check this propagation.

Nevertheless you should not state a hard coded 30 seconds without any way to modify it. For any reason, the provider could take more than 30 seconds to propagate then changes. So you should put in the provider options a parameter to modify this value.

And then, you have another solution to fasten offline tests: record the modifications normally with the 30 s time sleep, then push you code along with the cassettes by overriding the parameter to a low value, like 1s. During the CI, cassettes will be replayed correctly, but without waiting 30 s each time.

lexicon/providers/hetzner.py
rembik: That won't work here. The API responses with status code 200 and returns the unchanged zone file version.

I could look repeatedly until the changes are took over, but that came with a big overload... Any other idea?

adferrand: Yes, I added another proposition below.

rembik:
I think @adferrand and me came up with a stable version. What do you think of it, @AnalogJ ?

rembik:
Ok guys, I would love to see this provider merged to this great lib. What else do you need to make that happen @adferrand @AnalogJ?

AnalogJ:
Hey @rembik @adferrand 
Sorry about the delay, I've been traveling and didn't have time to take a look at this PR. It looks like @adferrand gave you a bunch of pointers though.

I'll try to take a look at this tonight/tomorrow and merge it if everything looks good.

AnalogJ:
Hey @rembik  can you add an entry in the README.md for this provider?

PR #328, opened 2018-12-04
Added SRV record handling to constellix

oliverisaac:
This adds the initial ability to add SRV records using constellix. Of note, all weights and priorities will be the same until that part of the code is patched.

### DIFF
--- a/lexicon/providers/constellix.py
+++ b/lexicon/providers/constellix.py
@@ -39,6 +39,8 @@ def ProviderParser(subparser):
         "--auth-username", help="specify the API key username for authentication")
     subparser.add_argument(
         "--auth-token", help="specify secret key for authenticate=")
+    subparser.add_argument("--weight", help="specify the SRV record weight")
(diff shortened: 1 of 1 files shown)


AnalogJ:
Thanks for the PR @oliverisaac! 

I completely understand why this change is necessary. When building `lexicon` most of my focus was on the popular/basic record types and options, and I didnt take into account all the additional configuration options available for some of them. 

While this PR would add support for SRV customization options for `constellix` I would prefer a more general purpose/flexible approach that would be usable by multiple providers. 

If you feel comfortable enough with the `lexicon` codebase and have a good idea how to add such a feature, I'd be happy to provide feedback on any PR you submit that implements a more generic solution. 

Otherwise, I'd appreciate it if you could create an issue so that we can track this for future development.

adferrand:
I am closing this PR, and looking forward to see a more general proposition from you @oliverisaac, if you have still the time!

PR #334, opened 2018-12-19
Added fallback handling for resolving DNS

rembik:
Added fallback handling for resolving DNS when no domain found via local nameservers. Solves issue #333

### DIFF
--- a/lexicon/providers/hetzner.py
+++ b/lexicon/providers/hetzner.py
@@ -8,6 +8,7 @@ import re
 import time
 import requests
 from six import string_types
+import tldextract
 from urllib3.util.retry import Retry
 
 # Due to optional requirement
@@ -423,6 +424,7 @@ class Provider(BaseProvider):
             resolver = dns.resolver.Resolver()
             if nameservers:
                 resolver.nameservers = nameservers
+            resolver.lifetime = 1
             rrset = resolver.query(name, rdtype)
             for rdata in rrset:
                 LOGGER.debug('DNS Lookup => %s %s %s %s',
@@ -459,7 +461,12 @@ class Provider(BaseProvider):
         more linked record name was found for the given fully qualified record name or
         the CNAME lookup was disabled, and then returns the parameters as a tuple.
         """
-        domain = dns.resolver.zone_for_name(name).to_text(True)
+        resolver = dns.resolver.Resolver()
+        resolver.lifetime = 1
+        try:
+            domain = dns.resolver.zone_for_name(name, resolver=resolver).to_text(True)
+        except dns.exception.DNSException:
+            domain = tldextract.extract(name).registered_domain
         nameservers = Provider._get_nameservers(domain)
         cname = None
         links, max_links = 0, 5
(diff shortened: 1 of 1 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/334)

---

rembik:
Hello @adferrand, first of all: dns.resolver does not need an internet connection, but needs a nameserver that knows the requested domain. That is the case in production environments.

Why tldextract is only fallback solution:
If the domain zone is equal to the 2nd-level domain, tldextract will do the job. However, this is not always the case and the reason why tldextract is just a fallback to dns.resolver for this particular provider.

This provider supports a CNAME lookup and record propagation check. For both options, we want the domain zone and nameservers to be correct so that the provider API can work as expected and fast.

adferrand:
Indeed. So that was my guess: the fallback is not a valid one, because it does not cover everything you need. You implement it because tests could fail without internet access.

But you should not add in the code you want a test a logic specifically designed to make the test succeed. Because in this case, the code to test behave differently whether you test it or execute it in production, so it is not a real test.

Your real problem is that when you call the dns module, it may fail because of the environment. I will add another thing to the initial problem. Even if the CI system has access to internet all the time, one day, the domain you used to write the tests will not be available anymore: so your test will start to fail, or will start to not correctly test because of the fallback.

So in fact, you need to fix this part of the tests environment, and that the purpose of mocking. Check the tests in the `auto` provider, I had exactly the same problem that you. The idea is to mock at the lowest level the method whose execution depends on the environment, and replace it by a mocked function, that will always return what you would expect in production.

For me, it was the function that calls `nslookup`. For you, I think it will be the exact method in dns module that raise a timeout.

Note that I also add a specific and unique test for the real method that otherwise is mocked: but the test is executed only if network is available, and for a domain that is likely to stay for a while (Google). But even if Google one day disappear, it will then only a matter of switching a single line of code for future maintainers, not a story of recording again all cassettes.

tests/providers/test_hetzner.py
```diff
@@ -30,13 +34,14 @@ def get_dns_cname_mock(self, request):
...
 
-    @pytest.mark.skipif(no_network(), reason='No network, no dns.resolver call possible.')
-    @pytest.mark.ignore_get_dns_cname_mock('yes')
+    @pytest.mark.skipif(_no_dns_lookup(), reason='No DNS resolution possible.')
+    @pytest.mark.ignore_dns_cname_mock
     def test_get_dns_cname(self):
         """Ensure that zone for name can be resolved through dns.resolver call."""
-        _domain, _, _ = Provider._get_dns_cname('_acme-challenge.fqdn.{}.'.format(self.domain),
-                                                False)
+        _domain, _nameservers, _ = Provider._get_dns_cname(('_acme-challenge.fqdn.{}.'
```
adferrand: Can we assert something about the third element returned by the tuple?

tests/providers/test_hetzner.py
```diff
@@ -1,25 +1,29 @@
...
-    try:
-        socket.create_connection(("www.google.com", 80))
-        return False
-    except OSError:
-        pass
-    return True
-
 class HetznerIntegrationTests(IntegrationTests):
 
+    def _no_dns_lookup():
```
adferrand: You should move this method outside the class. Or add self as stated by the pylint error.

adferrand:
@rembik, almost done! Your implementation for the tests is now correctly protected from unexpected runtime situations. Indeed testing the DNS resolution instead of testing the network connection is a good abstraction level in this situation.

tests/providers/test_hetzner.py
rembik: Did not find the time to do it after the last push... as use for pytest skip condition, we only can move this method outside the class....

tests/providers/test_hetzner.py
rembik: If present, the third element is the CNAME. Otherwise None (see docstring). Because I use productiv test domains, it could not be sure if there will be a CNAME sometime. So no, we can not assert something.

adferrand: Fair enough. Mock could be done to ensure a particular situation (record is a cname or not), but I do not see a real value to do it.

adferrand: Oh, sorry I did not see you found a way!

PR #336, opened 2018-12-27
Improve & fix CLI output with non-list actions

### LINKED ISSUE #272: Memset provider: TypeError: string indices must be integers
Hi,

When using the Memset provider with the default table formatting I get this error:

```bash
$ lexicon memset create example.com TXT --name _acme-challenge.example.com --content BLAH --ttl 300
  File "/usr/local/bin/lexicon", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/lexicon/__main__.py", line 133, in main
    handle_output(results, parsed_args.output)
  File "/usr/local/lib/python2.7/dist-packages/lexicon/__main__.py", line 109, in handle_output
    table = generate_table_result(logger, results, output_type == 'TABLE-NO-HEADER')
  File "/usr/local/lib/python2.7/dist-packages/lexicon/__main__.py", line 75, in generate_table_result
    array = [[row['id'], row['type'], row['name'], row['content'], row['ttl']] for row in output]
TypeError: string indices must be integers
```

I think this is because `output` is a string not an array - when I added `print output` I got a string like `969f9caabe19859c11249333dd80aa15`.

When I use `--output JSON` I get the same ID plus quotes:

```bash
$ lexicon memset create example.com TXT --name _acme-challenge.example.com --content BLAH --ttl 300 --output JSON
"969f9caabe19859c11249333dd80aa15"
```

I know Memset's not public so if you need any help to test it just let me know. For now I'll work around it with `--output QUIET` since I don't really care about the output here.

Thanks!
Dave

adferrand:
This PR improves how output is handled for non-list (eg. create) actions. Indeed, current output handling is designed only to tread correctly dict results generated by list actions. For other actions, output will depend on the nature of the object return by the action, and the chosen output type.

Question is more critical for TABLE output type, that is the default output:
* for boolean/integer results, as theses objects are not iterable, output will be silently suppressed, as with QUIET output option
* for string results, as theses objects are iterable but are not typical dicts returned during a list action, it will lead to an exception.

This issue can be avoid currently by setting the output to QUIET or JSON.

In this PR, non-list action results are handled correctly, by displaying them as expected with TABLE, TABLE-NO-HEADER and JSON. Also for JSON, all objects (not only iterable objects) can be displayed now, if they are JSON serializable. Non serializable objects will lead to no output, not an exception.

Fixes #272

### DIFF
--- a/lexicon/cli.py
+++ b/lexicon/cli.py
@@ -14,12 +14,12 @@ from lexicon.parser import generate_cli_main_parser
 logger = logging.getLogger(__name__)  # pylint: disable=C0103
 
 
-def generate_table_result(lexicon_logger, output=None, without_header=None):
-    """Convert returned JSON into a nice table for command line usage"""
+def generate_list_table_result(lexicon_logger, output=None, without_header=None):
+    """Convert returned data from list actions into a nice table for command line usage"""
     try:
-        _ = (entry for entry in output)
+        isinstance(output, dict)
     except TypeError:
-        lexicon_logger.debug('Command output is not iterable, and then cannot '
+        lexicon_logger.debug('Command output is not a dict, and then cannot '
                              'be printed with --quiet parameter not enabled.')
         return None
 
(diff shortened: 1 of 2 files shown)


AnalogJ:
wow, you can definitely tell that I dont use the table outputs at all. This is great 👍

PR #338, opened 2018-12-29
[Minor] Full setuptools

adferrand:
This PR rationalizes dependencies handling in Lexicon.

Indeed currently dependencies are scattered among two mechanisms:
* runtime (mandatory + optional) dependencies are declared in `setup.py` as required to make a distributable package
* pip requirements files for runtime and test dependencies, with `requirement.txt` refering to an editable installation of lexicon (so implicitly `setup.py`), `optional-requirements.txt` listing also editable installation for each extra defined in `setup.py`, and test-requirements.txt that contains explicitly requirements for test/dev purpose.

The `*.txt` requirements files are removed, and respective dependencies are consolidated directly in `setup.py` using extra mechanism. Also, a special extra `full` allows to install every optional dependencies, to have a Lexicon instance able to run any supported provider.

As a consequence, expressing dependencies in various systems, like `tox`, is more straight-forward, as it follows the standard pattern of `setuptools`. Once this PR is distributed to users through pypi, one will be able to install a full Lexicon instance using `pip install dns-lexicon[full]`.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/338)

---

AnalogJ:
This change looks fine, however keep in mind that the `requirements.txt` fill will keep getting re-created by CapsuleCD until `mgr_keep_lock_file` is set to false in the `caspsule.yml`.

https://github.com/AnalogJ/capsulecd/blob/master/pkg/mgr/mgr_python_pip.go#L68

Something else (not directly related to your PR) is that with our move away from requirements.txt, any version constraints we add to `setup.py` will enforce version constraints on upstream projects that use `lexicon` as a library, like certbot.

adferrand:
Certbot handle anyway the constraints coming transitively from there dependencies, so we are fine with it. What we need to care about it not adding new dependencies, or rely on fonctionnalities on newer versions that Certbot do not support.

But we are free for the `test` extra for instance, and in theory for the other extras, including `full`. Because theses dependencies are propagated through dns plugins, that are handled separately.

I think anyway that it is better for maintainability, because we will setup more easily a test runtime that corresponds to the production runtime of the library when it is used in upstream projects.

AnalogJ:
Yep, I just wanted to make a note of it so that we're on the same page. 👍

adferrand:
No problem, I will keep that in mind. Not to mention that if I screw up, I will be put in the first line for fixes on Certbot side ^^

adferrand:
I added the change you suggested to capsulecd. Is it OK for you?

AnalogJ:
Yep, looks good. Thanks

PR #339, opened 2018-12-29
Refactor tests in a python package

adferrand:
A standard pattern for python unit tests is to treat them as standard python modules of the project, and put them in a dedicated `tests` subpackage. Doing so ensures in particular consistency in module import in the unit tests, and fits better for pylint and mypy executions.

This PR does that.

Fixtures stays at their place, outside the `lexicon` namespace, as assets. Otherwise, they would be integrated in the distribution package, increasing drastically its weight. Furthermore, cassettes are intended to be used locally using the git repository.

I also add the possibility to specify another cassettes paths using the environment variable `LEXICON_VCRPY_CASSETTES_PATH`.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/339)

---

lexicon/tests/providers/integration_tests.py
```diff
@@ -10,17 +9,20 @@
...
 
 
-# Configure VCR. Parameter record_mode depends on the LEXICON_LIVE_TESTS environment variable value.
-record_mode = 'new_episodes' if os.environ.get(
-    'LEXICON_LIVE_TESTS', 'false') == 'true' else 'none'
+# Configure VCR. Parameter record_mode depends
+# on the LEXICON_LIVE_TESTS environment variable value.
+record_mode = 'none'
+if os.environ.get('LEXICON_LIVE_TESTS', 'false') == 'true':
+    record_mode = 'new_episodes'
```
AnalogJ: I wonder if its possible to do a check of the existing Cassettes and warn the user if the files already exist. In general we want clean cassettes with only one set of recordings.

AnalogJ:
LGTM 👍

lexicon/tests/providers/integration_tests.py
adferrand: Do you mean with the `new_episodes` mode ? Because during the `none` mode, no cassette will be written, and if no cassette correspond to the test, or a cassette gives invalid request replay, an error will be thrown.
For `new_episodes`, if I recall correctly, the relevant cassette will be reused if it exists, and exceptions will be thrown also if the request replay does not correspond. If no cassette exists, then it is written. So with that approach, cassettes would not be rewritten without explicit action (deleting manually the cassette before).

adferrand: In fact, the behavior I describe is the `once` mode. The mode `new_episodes` will always rewrite the cassette. 

`once` became the default mode in recent versions of vcrpy. I think it would be relevant to do it also in lexicon.

adferrand:
Do you mean the `tests` directory at the root of Lexicon project? Well, it is still relevant that it exists for capsulecd, and for Lexicon in general. It is there that cassettes are stored, and I also for scripts that are not unit tests, but are part of the CI, like the quality gate.

So in a way or another, it is correct that this directory stay the standard test folder for capsulecd, and that it always exists.

lexicon/tests/providers/integration_tests.py
AnalogJ: yeah, thats what I meant, `once` is probably the correct solution here. During development a developer may be creating changes that modify the HTTP requests, and should be deleting the cassettes for their provider sporadically. Only the actually used HTTP requests should be included in the final PR.

AnalogJ:
Ah, I think I just mis-remembered then. That should be fine then, forget I said anything :)

lexicon/tests/providers/integration_tests.py
adferrand: All right, I pushed a commit to use once mode for live tests in this PR.

AnalogJ:
Hmm, looks like this PR needs to be rebased/merged with master.

adferrand:
Update. All systems green !

PR #342, opened 2019-01-06
Rename ProviderParser into provider_parser

adferrand:
As `ProviderParser` is a function, so it should be named `provider_parser`.

This PR modifies the name of this function in all provider modules. I did not put retro-compatibility logic on `ProviderParser` attribute name in the provider modules, because this function is for internal use, to build Lexicon CLI. In particulier Certbot uses only the `Provider` class, and nothing else.

But if it is a problem, I can build a retro-compatible logic.


AnalogJ:
nope, thats fine. LGTM 👍

PR #366, opened 2019-02-11
Improve support of optional dependencies at runtime and during tests

adferrand:
I recently discussed with the integrators of Lexicon for Debian. They explained to me that the fact some providers require some dependencies to work, is something quite complex to deal to build OS packages. Indeed, they execute the integration tests against available python dependencies, while skipping all optional dependencies to avoid packages with a complex graph.

However, this requires to maintain manually a list of providers that have optional dependencies, to avoid them and break the tests. They would like to have some way to do that out-of-the-box.

In parallel, I see that the CLI throws errors difficult to handle as a user when theses dependencies are missing. I would like to give some useful directions to the use in this case (typically, advice to install `lexicon[provider]`.

This PR responds to these needs:
* a new module, `lexicon.discovery`, handle the providers discovery and their availability towards current installed dependencies, using what the metadata generated by `setuptools` from the `extra` section of `setup.py`.
* a pytest CLI option, `--xfail-providers-with-missing-deps`, is available for test execution on the `lexicon` package (using `conftest.py`). When toggled, all providers with extra dependencies that are missing, will xfails. Xfail means that the test is expected to fail: if the test succeed, it is not normal and the pytest execution fails. I think it is more accurate than a simple skip, as it is a negative test.
* `tox` environment are updated: `standard` will execute all tests available, while supposing that all dependencies are installed, and `light` will trigger the `--xfail-providers-with-missing-deps`, and run with only the core dependencies.
* Lexicon will detect before loading effectively a module, if all dependencies are present. Otherwise, a "smart" `ProviderNotAvailableError` will be thrown with a useful message inviting to install relevant extra for this provider
* similarly, tests do not explode when the modules are loaded, by loading the relevant provider during the runtime, not during the import. I also take the occasion to simplify by this way imports and parameters of the tests classes.

Once landed on master, I will propose a mail to the integrators for Debian and Red Hat world to be sent.


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/366)

---

adferrand:
Hello @AnalogJ! Will you have some time to review this PR in the next days? I think it would improve a lot the clarity for users towards providers optional dependencies.

AnalogJ:
Hey @adferrand Sorry I had taken a look a while back, but I didnt end up merging this PR. let me do that now.

PR #399, opened 2019-04-16
Allow specifiying Gandi sharing ID

cnorthwood:
If your domain is in an organization within Gandi LiveDNS, but your API
token is a personal one, then you must specify the Sharing ID of the
organization which the domain you wish to modify is within.

### DIFF
--- a/lexicon/providers/gandi.py
+++ b/lexicon/providers/gandi.py
@@ -50,6 +50,10 @@ def provider_parser(subparser):
     subparser.add_argument(
         '--api-protocol',
         help="(optional) specify Gandi API protocol to use: rpc (default) or rest")
+    subparser.add_argument(
+        '--sharing-id',
+        help="Gandi sharing ID "
+             "(required if your domain is in an organization and using REST protocol)")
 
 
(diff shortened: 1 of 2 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/399)

---

AnalogJ:
This seems simple enough, but we don't have a test case that covers this at all. Could you add a pair of custom tests to the [provider test suite](https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/gandi.py) to validate the failing and passing API calls?

cnorthwood:
okay, upon further digging (as I couldn't get a test case to fail correctly)... it appears that the way this interacts with the Gandi API the sharing ID isn't needed, as it interacts directly with domain objects, rather than zone objects (which is where the sharing ID is needed) - and so assumes that a zone has already been created with the scope of the organisation and attached to a domain you already have write access to, so the sharing ID is inherited from manipulating the zone through the domain object, rather than the zone directly.

PR #401, opened 2019-04-18
Fix powerdns

primesoftware:
1) This PR fixes the inability to add records with the same name but different content. Lets Encrypt issues different challenge tokens for primary and wildcard domains, and looks for both TXT records when challenging. Numerous plugins using dns-lexicon fail because the first TXT record has its content replaced with the second when using an option like "-d mydomain.com,*.mydomain.com" for a cert.

2) Also added functionality to make master notify slaves of changes.

pan-net-security/certbot-dns-powerdns#1
robin-thoni/certbot-pdns#8

### DIFF
--- a/lexicon/providers/powerdns.py
+++ b/lexicon/providers/powerdns.py
@@ -185,21 +185,22 @@ class Provider(BaseProvider):
         for rrset in self.zone_data()['rrsets']:
             if rrset['type'] == rtype and self._fqdn_name(rrset['name']) == self._fqdn_name(name):
                 update_data = rrset
+
                 if 'comments' in update_data:
                     del update_data['comments']
-                update_data['changetype'] = 'REPLACE'
+
+                if content is None:
+                    update_data['records'] = []
+                    update_data['changetype'] = 'DELETE'
+                else:
+                    new_record_list = []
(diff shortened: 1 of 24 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/401)

---

AnalogJ:
This is great!
Can you also remove the following lines: https://github.com/AnalogJ/lexicon/blob/master/lexicon/tests/providers/test_powerdns.py#L27-L30

That'll enable the extended test suite which validates wildcards/record sets. 
You'll also need to add the recordings that are created.

primesoftware:
> 
> 
> This is great!
> Can you also remove the following lines: https://github.com/AnalogJ/lexicon/blob/master/lexicon/tests/providers/test_powerdns.py#L27-L30
> 
> That'll enable the extended test suite which validates wildcards/record sets.
> You'll also need to add the recordings that are created.

All pushed! I had to fix the delete record function too as that wasn't behaving properly either...

AnalogJ:
Awesome work, thanks for finishing this off :)

PR #406, opened 2019-04-30
add support for Aliyun

wildcloud3:
support for aliyun dns manipulation


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/406)

---

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+
+NAMESERVER_DOMAINS = ['hichina.com']
+
+ALIYUN_DNS_API_ENDPOINT = 'https://alidns.aliyuncs.com'
+ALIYUN_DNS_DEFAULT_REGION = 'cn-hangzhou'
+
+def provider_parser(subparser):
+    """Module provider for Aliyun"""
+    subparser.add_argument(
+        "--access-key-id", help="specify api key for authentication")
```
adferrand: Usually we use the convention `--auth-param` in Lexicon for params that contains sensible data for the provider authentication. Could `--access-key-id` be replaced by `--auth-key-id` ?

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = ['hichina.com']
+
+ALIYUN_DNS_API_ENDPOINT = 'https://alidns.aliyuncs.com'
+ALIYUN_DNS_DEFAULT_REGION = 'cn-hangzhou'
+
+def provider_parser(subparser):
```
adferrand: A good practice is also to fill the `description` of `subparser`. It allows to have a general description of the provider that is displayed when `lexicon provider --help` is invoked. It is a good place, for instance, to have a general description of the authentication used by the provider, and so what are the necessary sensible data that needs to be gathered and how. See for instance the provider `ovh` or `zeit`.

This is not mandatory at all, so if you consider that it does not worth it, I will not block on it.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+        "--access-key-id", help="specify api key for authentication")
+    subparser.add_argument(
+        "--access-secret", help="specify api secret for authentication")
+
+
+class Provider(BaseProvider):
+    """Provider class for Linode"""
+    def __init__(self, config):
+        super(Provider, self).__init__(config)
+        self.client = None
```
adferrand: Except if you want `client` to be exposed outside of the provider, you should mark it as private by renaming it `_client`.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _authenticate(self):
+        response = self._request('DescribeDomainInfo')
+
+        if 'DomainId' not in response:
+            raise Exception(response)
```
adferrand: Raise a more specific exception here, like `ValueError`.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+                'id': resource['RecordId'],
+                'type': resource['Type'],
+                'name': self._full_name(resource['RR']),
+                'ttl': resource['TTL'],
+                'content': resource['Value']
+            })
+        LOGGER.debug('list_records: %s', processed_records)
+        return processed_records
+
+    def __get_record(self, identifier):
```
adferrand: Maybe you want to name this method `_get_record`, with only one underscore ?

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+                and (self._relative_name(name) == self._relative_name(record['name'])) \
+                and (content == record['content']):
+                return True
+
+        if not identifier:
+            record = resources[0] if resources else None
+            identifier = record['id'] if record else None
+
+        if not identifier:
+            self._create_record(rtype, name, content)
```
adferrand: The contract of `update_record`, as specified in https://github.com/AnalogJ/lexicon/blob/master/SPECIFICATION.md#update_record, is to update an existing record, not to create it if it does not exist. You should throw an exception if no record to update could be found.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+            'id': response['RecordId'],
+            'type': response['Type'],
+            'name': self._full_name(response['RR']),
+            'ttl': response['TTL'],
+            'content': response['Value']
+        }
+
+    # Create or update a record.
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        resources = self._list_records(rtype, name, None)
```
adferrand: There could have more than once record that matches a given type and name (like for TXT entries for instance). What I do usually in this case is to display a warning saying that only the first one record will be updated. An other possibility is to raise an Error. Personally I prefer the warning.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+            delete_resource_id.append(identifier)
+
+        LOGGER.debug('delete_records: %s', delete_resource_id)
+
+        for resource_id in delete_resource_id:
+            self._request('DeleteDomainRecord', query_params={'RecordId': resource_id})
+
+        return True
+
+    def _request(self, action='GET', url='/', data=None, query_params=None):
```
adferrand: So to follow the API specifications, you need to significantly modify the inherited `_request` behavior, by passing a non-HTTP verb in the `action`, and ignoring `url`.

For the sake of code readiness, I think you should write your own specific request method with a different name, and just not use at all the inherited `_request` method.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+            query_params = {}
+
+        query_params.update(self._build_default_query_params(action))
+        query_params.update(self._build_signature_parameters())
+
+        query_params.update({'Signature': self._calculate_signature('GET', query_params)})
+
+        response = requests.request('GET', url, params=query_params)
+
+        # TODO: add exception handle
```
adferrand: Missing resolved TODO ? You can use `response.raise_for_status()` here for instance, if the errors are passed through the API using the HTTP status codes.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+        return response.json()
+
+    # @param query_params should be {} and not none
```
adferrand: You can check early for this statement in `_calculate_signature`, and raise an exception if it is not respected.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+    def _calculate_signature(self, http_method, query_params):
+        access_secret = self._get_provider_option('access_secret')
```
adferrand: `access_secret` may be `None` if the user forgot to pass the flag. It is reasonable to raise an exception here in case of `None` value, and avoid to send requests that have no chance to be authenticated.

lexicon/providers/aliyun.py
```diff
@@ -0,0 +1,217 @@
...
+            sign = hmac.new(sign_secret_bytes, string_to_sign_bytes, sha1)
+            signature = base64.b64encode(sign.digest()).decode()
+        else:
+            sign = hmac.new(sign_secret, string_to_sign, sha1)
+            signature = sign.digest().encode('base64').rstrip('\n')
+
+        return signature
+
+    def _build_signature_parameters(self):
+        access_key_id = self._get_provider_option('access_key_id')
```
adferrand: Here also, `access_key_id` may be `None`, raising an exception in this case will avoid to build falsy signatures parameters.

lexicon/tests/providers/test_aliyun.py
```diff
@@ -0,0 +1,17 @@
...
+
+# Hook into testing framework by inheriting unittest.TestCase and reuse
+# the tests which *each and every* implementation of the interface must
+# pass, by inheritance from integration_tests.IntegrationTests
+class AliyunProviderTests(TestCase, IntegrationTests):
+    """Integration tests for Foo provider"""
+    provider_name = 'aliyun'
+    domain = 'mean.space'
+
+    """filter access key id"""
```
adferrand: You should move this docstring just below `_filter_query_parameter` declaration, or change it as an inline comment prefixed by `#`.

lexicon/tests/providers/test_aliyun.py
```diff
@@ -0,0 +1,17 @@
...
+    def _filter_query_parameters(self):
+        return ['AccessKeyId', 'Signature', 'SignatureNonce', 'Timestamp']
```
adferrand: Not mandatory at all, but nice feature if you want to keep cassettes with sanitized sensible material that are not removed, just replaced by dummy values. VCRpy allows you to declare tuples of two elements (eg. `('AccessKeyId', 'KEY_ID')` instead of just the string of the query parameter name to filter (eg. `AccessKeyId`).

In this case, instead of just removing the parameter, VCRpy will replace its value in the cassette by the second value of the tuple.

https://vcrpy.readthedocs.io/en/latest/advanced.html#advanced-use-of-filter-headers-filter-query-parameters-and-filter-post-data-parameters

setup.py
```diff
@@ -40,7 +40,7 @@
     'henet': ['beautifulsoup4'],
     'hetzner': ['dnspython>=1.15.0','beautifulsoup4'],
     'easyname': ['beautifulsoup4'],
-    'localzone': ['localzone'],
```
adferrand: It is not necessary to remove the comma here. Python supports arrays with trailing comma, and it allows to have cleaner diff in GIT when a new value is added: indeed, instead of having two lines modified (the new one, and the previous one with an added comma), only one line is modified, the one that is added.

wildcloud3:
Thanks @adferrand 

I've done some optimization along to your advices.

PR #411, opened 2019-06-03
Provide SafeDNS provider integration 

miff2000:
This PR is to add in support for SafeDNS, UKFast's DNS service.

Let me know if I've missed anything :)

### DIFF
--- a/lexicon/providers/safedns.py
+++ b/lexicon/providers/safedns.py
@@ -50,12 +50,12 @@ class Provider(BaseProvider):
 
             # Assign the returned attributes to the keys that lexicon expects
             for record in payload['data']:
+                record = self._clean_TXT_record(record)
                 processed_record = {
                     'id': record['id'],
                     'name': record['name'],
                     'type': record['type'],
-                    # Make sure TXT records are wrapped before sending
-                    'content': self._remove_quotes(record['type'], record['content']),
+                    'content': record['content'],
                     'updated_at': record['updated_at'],
                     'ttl': record['ttl'],
                     'priority': record['priority']
@@ -184,12 +184,4 @@ class Provider(BaseProvider):
         if rtype == 'TXT':
             if not content.startswith('"') and not content.endswith('"'):
                 return '"{0}"'.format(content)
-        return content
-
-    # When we're getting a response, we don't want quotes to wrap the content.
-    # This method removes them
-    @staticmethod
-    def _remove_quotes(rtype, content):
-        if rtype == 'TXT':
-            return content.strip('"')
         return content
--- a/lexicon/providers/safedns.py
+++ b/lexicon/providers/safedns.py
@@ -105,6 +105,13 @@ class Provider(BaseProvider):
         return True
 
     def _update_record(self, identifier, rtype=None, name=None, content=None):
+
+        # Make sure the update won't cause a duplicate entry. If it will, fail silently
+        records = self._list_records(rtype, name, content)
+        if len(records) >= 1:
+            LOGGER.debug('update_record: (ignored, update would cause a duplicate record): $s', records[0]['id'])
+            return True
+        
         if not identifier:
             records = self._list_records(rtype, name)
             if len(records) == 1:
--- a/lexicon/providers/safedns.py
+++ b/lexicon/providers/safedns.py
@@ -109,9 +109,10 @@ class Provider(BaseProvider):
         # Make sure the update won't cause a duplicate entry. If it will, fail silently
         records = self._list_records(rtype, name, content)
         if len(records) >= 1:
-            LOGGER.debug('update_record: (ignored, update would cause a duplicate record): $s', records[0]['id'])
+            LOGGER.debug('update_record: (ignored, update would cause a duplicate record): %s',
+                         records[0]['id'])
             return True
-        
+
         if not identifier:
             records = self._list_records(rtype, name)
             if len(records) == 1:
--- a/lexicon/providers/safedns.py
+++ b/lexicon/providers/safedns.py
@@ -145,11 +145,6 @@ class Provider(BaseProvider):
         LOGGER.debug('delete_record: %s', True)
         return True
 
-    # As there is no _patch method available from Provider, let's create our own
-    # by creating a _request with action 'PATCH'.
-    def _patch(self, url='/', data=None, query_params=None):
-        return self._request('PATCH', url, data=data, query_params=query_params)
-
     def _request(self, action='GET', url='/', data=None, query_params=None):
         if data is None:
             data = {}
--- a/lexicon/providers/safedns.py
+++ b/lexicon/providers/safedns.py
@@ -56,15 +56,16 @@ class Provider(BaseProvider):
                     'id': record['id'],
                     'name': record['name'],
                     'type': record['type'],
-                    'content': record['content'],
+                    # Make sure TXT records are wrapped before sending
+                    'content': self._remove_quotes(record['type'],record['content']),
                     'updated_at': record['updated_at'],
                     'ttl': record['ttl'],
                     'priority': record['priority']
                 }
                 records.append(processed_record)
 
-        # This appears to be filtering logic to return only the record
-        # which matches what's passed in to the method
+        # This is filtering logic to return only the record which matches what has 
+        # been passed in to the method
         if rtype:
             records = [
                 record for record in records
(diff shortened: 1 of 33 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/411)

---

AnalogJ:
Hey @miff2000 thanks for the PR! 🎉 

@adferrand and I will take a look, and get back to you if we find anything that needs to be changed. If not you're PR should be included in the next release.

lexicon/providers/safedns.py
```diff
@@ -0,0 +1,200 @@
...
+            'name': self._full_name(name),
+            'type': rtype,
+            'content': content
+        }
+
+        self._post('/zones/{0}/records'.format(self.domain), data)
+        LOGGER.debug('create_record: %s', True)
+        return True
+
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
```
adferrand: There is one case you did not implement here, which is related to the integration test `test_provider_when_calling_update_record_should_modify_record_name_specified`. It is when Lexicon is invoked to update a record, but no identifier is passed. In this case, the `--name` flag must be passed, and `name` and `rtype` are used together to find the record.

I usually reuse ̀self.list_records` for that. If there is no record found, then there is no update. If at least one record is found, the first one is updated (with a warning if there is more than one record).

Once you have implemented that, don't forget to remove `test_provider_when_calling_update_record_should_modify_record_name_specified` from the skipped tests, and fill its associated recording.

lexicon/providers/safedns.py
```diff
@@ -0,0 +1,200 @@
...
+        for delete_record_id in delete_record_ids:
+            self._delete(
+                '/zones/{0}/records/{1}'.format(self.domain, delete_record_id))
+
+        LOGGER.debug('delete_record: %s', True)
+        return True
+
+    # As there is no _patch method available from Provider, let's create our own
+    # by creating a _request with action 'PATCH'.
+    def _patch(self, url='/', data=None, query_params=None):
```
adferrand: I think you can move this method to the base provider, so everyone can benefit from it.

lexicon/providers/safedns.py
```diff
@@ -0,0 +1,200 @@
...
+    def _add_quotes(rtype, content):
+        if rtype == 'TXT':
+            if not content.startswith('"') and not content.endswith('"'):
+                return '"{0}"'.format(content)
+        return content
+
+    # When we're getting a response, we don't want quotes to wrap the content.
+    # This method removes them
+    @staticmethod
+    def _remove_quotes(rtype, content):
```
adferrand: Maybe you can reuse `_clean_TXT_record` from the base provider instead ?

lexicon/tests/providers/test_safedns.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @pytest.mark.skip(reason="Record-level TTLs are not supported by this provider")
+    def test_provider_when_calling_list_records_after_setting_ttl(self):
+        return
+
+    @pytest.mark.skip(reason="CNAME requires FQDN for this provider")
+    def test_provider_when_calling_create_record_for_CNAME_with_valid_name_and_content(self):
+        return
+
+    @pytest.mark.skip(reason="new test, missing recording")
```
adferrand: See my comment on `update_record` function in the provider.

lexicon/tests/providers/test_safedns.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_provider_when_calling_update_record_should_modify_record_name_specified(self):
+        return
+
+    def _filter_response(self, response):
```
adferrand: This override is not necessary.

lexicon/providers/safedns.py
miff2000: Yep, good shout

lexicon/providers/safedns.py
miff2000: I didn't spot that in the base class. Definitely!

lexicon/tests/providers/test_safedns.py
miff2000: Removed

lexicon/providers/safedns.py
miff2000: I saw your logic in the ovh provider and I've mirrored similar in mine. Hopefully it satisfies.

miff2000:
One question @adferrand. In my setup I find that the test `test_provider_when_calling_update_record_should_modify_record` was causing duplicate TXT records to be created. Did you find that also? I've added logic now to silently fail if that happens. Is that the right behaviour?

adferrand:
> One question @adferrand. In my setup I find that the test `test_provider_when_calling_update_record_should_modify_record` was causing duplicate TXT records to be created. Did you find that also? I've added logic now to silently fail if that happens. Is that the right behaviour?

In fact you see a duplicate record because you launched (at least) two times the integration tests, without removing the created records between two consecutive iterations. Yes, these integration tests do not clean up things after execution, this is really something we should implement. This is not that obvious however, since if the clean up is relying on `delete_record` is broken at some point, you could end with a ineffective cleanup, or worse, an accidental wipe up of all the zone...

Nonetheless, you are seeing the records created by the previous iterations. So it is not necessary to put a condition for that. It may be counterproductive in fact, because it will make update_record return True, while no update has been done.

miff2000:
> > One question @adferrand. In my setup I find that the test `test_provider_when_calling_update_record_should_modify_record` was causing duplicate TXT records to be created. Did you find that also? I've added logic now to silently fail if that happens. Is that the right behaviour?
> 
> In fact you see a duplicate record because you launched (at least) two times the integration tests, without removing the created records between two consecutive iterations. Yes, these integration tests do not clean up things after execution, this is really something we should implement. This is not that obvious however, since if the clean up is relying on `delete_record` and it is broken at some point, you could end with an ineffective cleanup, or worse, an accidental wipe of all the zone...
> 
> Nonetheless, you are seeing the records created by the previous iterations. So it is not necessary to put a condition for that and ignore it. It may be counterproductive in fact, because it will make update_record return True, while no update has been done.

Oh okay, that explains it. I wrongly assumed that the tests were intended to be able to be run without manual cleanup. So, with that in mind, this is the logic that I've borrowed (😉) from your OVH provider:
```python
            records = self._list_records(rtype, name)
            if len(records) == 1:
                identifier = records[0]['id']
            elif len(records) > 1:
                raise Exception('Several record identifiers match the request')
            else:
                raise Exception('Record identifier could not be found')
```
In your comment [above](https://github.com/AnalogJ/lexicon/pull/411#discussion_r294514983) you said to update the first record and warn. The two differ of course, so would you recommend that I update I raise an exception on the 0 records found, and update the first only and output a warning to LOGGER.debug if more >1 records found?

miff2000:
Hey @adferrand, did you get chance to review these changes I made?

adferrand:
Doing it today!

adferrand:
About your question, indeed the behavior is not consistent between the providers. Including the ones I implemented, since we did not cleared that point in the spec. More generally, there were some discussions to make a more consistent API around update operations.

Since most recent providers update first matching record with a warning if there is more than one, I would do the update instead of raising an exception.

miff2000:
> About your question, indeed the behavior is not consistent between the providers. Including the ones I implemented, since we did not cleared that point in the spec. More generally, there were some discussions to make a more consistent API around update operations.
> 
> Since most recent providers update first matching record with a warning if there is more than one, I would do the update instead of raising an exception.

Okay great, thanks. I'm switching to log instead as you've suggested and I'll push the changes as soon as I'm done. Shouldn't be long

miff2000:
Pushed @adferrand and all CI steps except AppVeyor have passed. Hopefully we're ready to merge! 🙂

lexicon/providers/safedns.py
```diff
@@ -0,0 +1,206 @@
...
+        return True
+
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+
+        # Make sure the update won't cause a duplicate entry. If it will, fail silently
+        records = self._list_records(rtype, name, content)
+        if len(records) >= 1:
+            LOGGER.debug('update_record: (ignored, update would cause a duplicate record): %s',
+                         records[0]['id'])
+            return True
```
adferrand: Just one last thing, I think you can remove this logic now.

miff2000: Yes, annoyed I missed that. Corrected and pushed

PR #412, opened 2019-06-06
Dreamhost

chhsiao1981:
Add a provider for [Dreamhost DNS API](https://help.dreamhost.com/hc/en-us/articles/217560167-API_overview)

### DIFF
--- a/lexicon/providers/dreamhost.py
+++ b/lexicon/providers/dreamhost.py
@@ -1,0 +1,181 @@
+"""Module provider for Dreamhost"""
+from __future__ import absolute_import
+import json
+import logging
+import time
+
+import requests
+from lexicon.providers.base import Provider as BaseProvider
+
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = ['dreamhost.com']
+
+
+def provider_parser(subparser):
+    """Module provider for Linode"""
+    subparser.add_argument(
+        "--auth-token", help="specify api key for authentication")
+
+
+class Provider(BaseProvider):
+    """Provider class for Linode"""
+
+    def __init__(self, config):
+        super(Provider, self).__init__(config)
+        self.domain_id = None
+        self.api_endpoint = 'https://api.dreamhost.com/'
+
+    def _authenticate(self):
+        self.domain_id = None
+        payload = self._get('domain-list_domains')
+        data = payload.get('data', None)
+        if data is None:
+            raise Exception('Domain not found')
+
+        for domain in data:
+            if domain.get('domain', '') == self.domain:
+                self.domain_id = self.domain
+        if self.domain_id is None:
+            raise Exception('Domain not found')
+
+    def _create_record(self, rtype, name, content):
+        name = self._full_name(name)
+        payload = self._get('dns-add_record', query_params={
+            'record': name,
+            'type': rtype,
+            'value': content,
+        })
+
+        if (payload.get('result', '') != 'success' and
+                payload.get('data', '') != 'record_already_exists_remove_first' and
+                payload.get('data', '') != 'record_already_exists_not_editable'):
+            raise Exception('unable to add record: %s' % payload)
+
+        return True
+
+    # List all records. Return an empty list if no records found
+    # type, name and content are used to filter records.
+    # If possible filter during the query, otherwise filter after response is received.
+    def _list_records(self, rtype=None, name=None, content=None):
+
+        payload = self._get('dns-list_records')
+
+        if payload.get('result', '') != 'success':
+            raise Exception('unable to get records: %s' % payload)
+
+        resource_list = payload.get('data', None)
+        if resource_list is None:
+            raise Exception('unable to get records: %s' % payload)
+
+        resource_list = [
+            resource for resource in resource_list if resource['zone'] == self.domain]
+        if rtype:
+            resource_list = [
+                resource for resource in resource_list if resource['type'] == rtype]
+        if name:
+            name = self._full_name(name)
+            resource_list = [
+                resource for resource in resource_list if resource['record'] == name]
+        if content:
+            resource_list = [
+                resource for resource in resource_list if resource['value'] == content]
+
+        processed_records = []
+        for resource in resource_list:
+            processed_records.append({
+                'id': resource['type'] + '|' + resource['record'] + '|' + resource['value'],
+                'type': resource['type'],
+                'name': resource['record'],
+                'content': resource['value'],
+            })
+
+        return processed_records
+
+    # Create or update a record.
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        if identifier == '':
+            identifier = None
+        if identifier is not None:
+            id_list = identifier.split('|')
+            if len(id_list) != 3:
+                raise Exception('invalid identifier: %s' % (identifier))
+
+            id_type, id_name, id_content = id_list
+
+            try:
+                self._delete_record(rtype=id_type, name=id_name, content=id_content)
+            except:  # pylint: disable=bare-except
+                pass
+
+        return self._create_record(rtype=rtype, name=name, content=content)
+
+    # Delete an existing record.
+    # If record does not exist, do nothing.
+    def _delete_record(self, identifier=None, rtype=None, name=None, content=None):
+
+        id_type = None
+        id_name = None
+        id_content = None
+        if identifier is not None:
+            id_list = identifier.split('|')
+            if len(id_list) != 3:
+                raise Exception('invalid identifier: %s' % (identifier))
+
+            id_type, id_name, id_content = id_list
+
+        if rtype is None:
+            rtype = id_type
+        if name is None:
+            name = id_name
+        if content is None:
+            content = id_content
+
+        if name is not None:
+            name = self._full_name(name)
+
+        to_deletes = []
+        if rtype is None or name is None or content is None:
+            if identifier is None:
+                records = self._list_records(rtype=rtype, name=name, content=content)
+                to_deletes = records
+        else:
+            to_deletes = [{'record': name, "type": rtype, "value": content}]
+
+        for each in to_deletes:
+            try:
+                self._get('dns-remove_record', query_params=each)
+            except:  # pylint: disable=bare-except
+                pass
+
+        return True
+
+    # Helpers
+    def _request(self, action='GET', url='', data=None, query_params=None):
+        time.sleep(1)
+        if data is None:
+            data = {}
+
+        if query_params is None:
+            query_params = {}
+
+        default_headers = {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json'
+        }
+
+        query_params['api_key'] = self._get_provider_option('auth_token')
+        query_params['format'] = 'json'
+        if 'cmd' not in query_params:
+            query_params['cmd'] = url
+
+        response = requests.request(action, self.api_endpoint, params=query_params,
+                                    data=json.dumps(data),
+                                    headers=default_headers)
+        # if the request fails for any reason, throw an error.
+        response.raise_for_status()
+        result = response.json()
+        if result.get('result', '') != 'success':
+            raise Exception('Dreamhost api error: {0}'.format(result))
+        return result
--- a/lexicon/providers/dreamhost.py
+++ b/lexicon/providers/dreamhost.py
@@ -1,5 +1,6 @@
 """Module provider for Dreamhost"""
 from __future__ import absolute_import
+import base64
 import json
 import logging
 import time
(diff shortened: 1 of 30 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/412)

---

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+from lexicon.providers.base import Provider as BaseProvider
+
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = ['dreamhost.com']
+
+
+def provider_parser(subparser):
+    """Module provider for Linode"""
```
adferrand: You should replace Linode for Dreamhost here.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+    subparser.add_argument(
+        "--auth-token", help="specify api key for authentication")
+
+
+class Provider(BaseProvider):
+    """Provider class for Linode"""
```
adferrand: You should replace Linode for Dreamhost here.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+            resource_list = [
+                resource for resource in resource_list if resource['record'] == name]
+        if content:
+            resource_list = [
+                resource for resource in resource_list if resource['value'] == content]
+
+        processed_records = []
+        for resource in resource_list:
+            processed_records.append({
+                'id': resource['type'] + '|' + resource['record'] + '|' + resource['value'],
```
adferrand: If you want, you can reuse the method `_identifier` from godaddy provider, to build a more ID-like value here.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+                'type': resource['type'],
+                'name': resource['record'],
+                'content': resource['value'],
+            })
+
+        return processed_records
+
+    # Create or update a record.
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        if identifier == '':
```
adferrand: Here, you can simply make a `if not identifier ... else ...` since an empty string, like an empty array, a `None` or a `False`, are falsy values.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+            identifier = None
+        if identifier is not None:
+            id_list = identifier.split('|')
```
adferrand: I see your smart move here, that avoids to request again the server and directly get the values you want from the identifier.

However, I would prefer a "real" ID (meaning something more like a hash, like in the `_identifier` I mentionned above, for two reasons:
* from a development point of view, this is brittle: for instance, if you have a `|` in your TXT content, you go in the gray zone with potential side-effects
* from a user point of view, this kind of concatenation will give not very handful ID to manipulate. For instance, a very long TXT content will give a very long ID. Carriage returns in a TXT value or weird characters will make it worse.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+            if len(id_list) != 3:
+                raise Exception('invalid identifier: %s' % (identifier))
+
+            id_type, id_name, id_content = id_list
+
+            try:
+                self._delete_record(rtype=id_type, name=id_name, content=id_content)
+            except:  # pylint: disable=bare-except
+                pass
```
adferrand: Can you put a comment here, to explain what is the expected exception, and why it is not a problem ?

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+
+    # Delete an existing record.
+    # If record does not exist, do nothing.
+    def _delete_record(self, identifier=None, rtype=None, name=None, content=None):
+
+        id_type = None
+        id_name = None
+        id_content = None
+        if identifier is not None:
+            id_list = identifier.split('|')
```
adferrand: Same as above,  in `update_record`, about the usage of a concatenation as an ID value.

lexicon/tests/providers/test_dreamhost.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Hook into testing framework by inheriting unittest.TestCase and reuse
+# the tests which *each and every* implementation of the interface must
+# pass, by inheritance from integration_tests.IntegrationTests
+class DreamhostProviderTests(TestCase, IntegrationTests):
+    """TestCase for Linode"""
+    provider_name = 'dreamhost'
+    domain = 'lexicon-example.com'
+
+    def _filter_post_data_parameters(self):
```
adferrand: You can remove this method is there is nothing to filter in post data.

lexicon/tests/providers/test_dreamhost.py
```diff
@@ -0,0 +1,27 @@
...
+        return []
+
+    def _filter_headers(self):
```
adferrand: You can remove this method if there is not header to remove.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+            if len(id_list) != 3:
+                raise Exception('invalid identifier: %s' % (identifier))
+
+            id_type, id_name, id_content = id_list
```
adferrand: These variables are not used after. What are their purpose?

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+                records = self._list_records(rtype=rtype, name=name, content=content)
+                to_deletes = records
+        else:
+            to_deletes = [{'record': name, "type": rtype, "value": content}]
+
+        for each in to_deletes:
+            try:
+                self._get('dns-remove_record', query_params=each)
+            except:  # pylint: disable=bare-except
+                pass
```
adferrand: Can you explain what is the expected exception here, and why it is not a problem ?

lexicon/tests/providers/test_dreamhost.py
```diff
@@ -0,0 +1,27 @@
...
+
+import pytest
+from lexicon.tests.providers.integration_tests import IntegrationTests
+
+
+# Hook into testing framework by inheriting unittest.TestCase and reuse
+# the tests which *each and every* implementation of the interface must
+# pass, by inheritance from integration_tests.IntegrationTests
+class DreamhostProviderTests(TestCase, IntegrationTests):
+    """TestCase for Linode"""
```
adferrand: You should update you comment to put Dreamhost instead of Linode.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+        id_name = None
+        id_content = None
+        if identifier is not None:
+            id_list = identifier.split('|')
+            if len(id_list) != 3:
+                raise Exception('invalid identifier: %s' % (identifier))
+
+            id_type, id_name, id_content = id_list
+
+        if rtype is None:
```
adferrand: Here, and for the next to `if` branches, you potentially end with a mix of values extracted from the id, and from the values passes explicitly by `--name` and `--content` CLI flags.

It should be one or the other. Either everything is taken from `identifier` if it is set, or we fallback to given `rtype`, `name` and `content`.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+            name = id_name
+        if content is None:
+            content = id_content
+
+        if name is not None:
+            name = self._full_name(name)
+
+        to_deletes = []
+        if rtype is None or name is None or content is None:
+            if identifier is None:
```
adferrand: I think we can improve in general the clarity of the branching in the `delete_record` method, because I think it becomes a little difficult to understand in which state we are at this point. By making a high level if condition on top of ` delete_record` to test the nullity of `identifier`, you would have two clear and separated logic for the two situations: if `identifier` is set, and if it is not. See also with my proposition about using a hash-based ID, that will influence the design of `delete_record`.

lexicon/providers/dreamhost.py
```diff
@@ -0,0 +1,181 @@
...
+            try:
+                self._get('dns-remove_record', query_params=each)
+            except:  # pylint: disable=bare-except
+                pass
+
+        return True
+
+    # Helpers
+    def _request(self, action='GET', url='', data=None, query_params=None):
+        time.sleep(1)
```
adferrand: Why do we need to wait here ?

chhsiao1981: This intends to avoid rate-limit or DDOS to Dreamhost.
There is a for-loop in _delete_record.

lexicon/providers/dreamhost.py
chhsiao1981: I'll replace to Dreamhost.

lexicon/providers/dreamhost.py
chhsiao1981: I'll replace to Dreamhost.

lexicon/providers/dreamhost.py
chhsiao1981: This is "trying to delete old record" in _update_record.
It's possible that the exception is because record is already deleted / does not exist.

I thought the focus of _update_record is to create the expected record,
and it's ok to ignore the exception in _delete_record.

Of course we can fine-tune this part to restrict except to only "non-exist".

lexicon/providers/dreamhost.py
adferrand: Fair enough! In this case, could you move the sleep up to the loop in delete method, if the other ones do not need to sleep, and add a comment explaining this is to avoid trigerring ddos protecting in case of looped requests?

lexicon/tests/providers/test_dreamhost.py
chhsiao1981: I'll remove this method.

lexicon/tests/providers/test_dreamhost.py
chhsiao1981: I'll remove this method.

lexicon/providers/dreamhost.py
adferrand: In this case yes I think we should focus on ignoring the request only if an http error occurred because of non-existant record. The other unexpected exceptions should stop the process right here.

lexicon/providers/dreamhost.py
chhsiao1981: I am not sure about the policy that what if we successfully delete 1 record but fail 1 another record and there are 2 remaining to-delete-records in the for-loop.
(How to deal with the 2 remaining to-delete-records?)

I think I'll change to the following setup:
1. for-loop all the candidate-to-delete-record
2.      catch the exception as err.
3. after for-loop, if err is not None: raise err (catch the last exception)

lexicon/tests/providers/test_dreamhost.py
chhsiao1981: I'll change to Dreamhost.

lexicon/providers/dreamhost.py
chhsiao1981: The problem with dreamhost is that they provide a very primitive "dns-list_records" api
in that this api just dumps all the records ACROSS ALL DOMAINS that you own.
It means that it's not efficient to just get the list of records and do the filtering.

When deleting a record, dreamhost also requires that we need to provide the value to delete.
It means that we do need the value-information to let the users be able to update / delete the record.

I admit that "|" separation is a very bad implementation.
Though it's getting longer, I think I'll wrap the id with json.dumps({"type": type, "record": record, "value": value}).

lexicon/providers/dreamhost.py
chhsiao1981: I'll simplify with "if not identifier:"

lexicon/providers/dreamhost.py
chhsiao1981: I'll wrap with Provider._identifier to make the code cleaner.

However, as explained in the later sections, I think I'll use json.dumps as the core-function in Provider._identifier.

lexicon/providers/dreamhost.py
chhsiao1981: I'll change to either we delete based on identifier if identifier is set, or we fallback to list the records based on rtype, name, content and delete the records.

chhsiao1981:
hi @adferrand 

Thank you so much for reviewing this PR.
I appreciate it a lot.
(I do want to have this PR be merged,
 because our organization is using Dreamhost, and we would like to use certbot to do letsencrypt.)

I think the only issue left is the format of id.

As mentioned in the reply,
Dreamhost provides a very primitive "dns-list_records" api
in that this api just dumps all the records ACROSS ALL DOMAINS that you own.
It means that it's not efficient to just get the list of records and do the filtering.

When deleting a record, Dreamhost also requires that we need to provide the value to delete.
It means that we do need the value-information to let the users be able to update / delete the record.

I admit that "|" separation is a very bad implementation.
Though it's getting longer, I think I'll wrap the id with json.dumps({"type": type, "record": record, "value": value}).

adferrand:
Ah, I see. Indeed I understand your concern tonkot make a really heavy request for almost nothing. I agree with the JSON approach you propose, that will make it less brittle. On top of it, in order to avoid to manipulate weird characters from the command line, I propose you to base64 decode/encode the JSON string from/to the command line. This way it will less error prone to pass a base64 line than a json string.

lexicon/providers/dreamhost.py
adferrand: Yes it is a good approach.

adferrand:
Hello @chhsiao1981, I did not saw yet new commits that were asserting the conversations we had in this PR. Do you plan to do it soon ?

Thanks again for your contribution !

chhsiao1981:
hi @adferrand ～

Sorry for the late of the response～
Sorry for my recent busy schedule.
I do plan to finish new modification by the end of this week～

Thanks～

adferrand:
No problem !

chhsiao1981:
though the tests are whole mocked-up, I tested with my real domain name and passed all the tests.

adferrand:
One last thing one the conversation about time.sleep usage in _request, and we are good to go!

PR #414, opened 2019-06-21
Dinahosting provider

iperurena:
Add a provider for [Dinahosting DNS API](https://en.dinahosting.com/api).

### DIFF
--- a/lexicon/providers/dinahosting.py
+++ b/lexicon/providers/dinahosting.py
@@ -1,0 +1,13 @@
+"""Module provider for Dinahosting"""
+from __future__ import absolute_import
+#import re
+import hashlib
+import logging
+
+import requests
+from lexicon.providers.base import Provider as BaseProvider
+
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = ['dinahosting.com']
--- a/lexicon/providers/dinahosting.py
+++ b/lexicon/providers/dinahosting.py
@@ -15,6 +15,12 @@ NAMESERVER_DOMAINS = ['dinahosting.com']
 # These record types are either not supported by the API or are non-compliant
 UNSUPPORTED_TYPES = ['MX', 'SRV', 'NS', 'SOA', 'LOC']
 
+# The API returns the record's content in any of these keys
+CONTENT_KEYS = ['ip', 'address', 'text', 'destinationHostname']
+
+# The API returns the record's name in any of these keys
+NAME_KEYS = ['host', 'hostname']
+
 # API response codes
 RC_SUCCESS = 1000
 RC_SUCCESS_PENDING = 1001
@@ -67,7 +73,7 @@ class Provider(BaseProvider):
         self.domain_id = self.domain
 
     def _create_record(self, rtype, name, content):
-        self._check_unsupported_type(rtype)
+        Provider._check_unsupported_type(rtype)
 
         rel_name = self._relative_name(name)
         data = {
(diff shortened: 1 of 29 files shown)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/414)

---

lexicon/providers/dinahosting.py
```diff
@@ -0,0 +1,258 @@
...
+        else:
+            data = {
+                'command': 'Domain_Zone_GetAll',
+                'domain': self.domain
+            }
+
+        payload = self._post('', data)
+
+        #content_obj = self._parse_record_content(rtype, content)
+        #content = content_obj['data']
```
adferrand: Some commented code to remove.

lexicon/providers/dinahosting.py
```diff
@@ -0,0 +1,258 @@
...
+                'id': hashlib.md5((ptype+full_name+raw_content).encode('utf-8')).hexdigest()
+            }
+            records.append(processed_record)
+
+        LOGGER.debug('list_records: %s', records)
+        return records
+
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        if identifier:
+            records = self._list_records(rtype)
```
adferrand: If this is not too heavy in term of API call, I think you could just get all records in the zone, not filtering on the `rtype`, then find the matching `identifier`. It allows a slightly wider usage of `update_record`.

lexicon/providers/dinahosting.py
```diff
@@ -0,0 +1,258 @@
...
+            matching_records = [record for record in records if record['id'] == identifier]
+        else:
+            matching_records = self._list_records(rtype, name)
+
+        # Check if record exists, otherwise raise an error
+        if not matching_records:
+            raise Exception("No matching records found. Cannot update.")
+        if len(matching_records) > 1:
+            raise Exception("Multiple matching records found. Cannot update.")
```
adferrand: It is your choice here, since we did not define precisely the spec when multiple records are encountered during an update. My current choice, on recent providers, is to take the first one, and log.warn about the fact that multiple records have been found.

Your choice here :)

lexicon/providers/dinahosting.py
```diff
@@ -0,0 +1,258 @@
...
+        delete_result = self._delete_record(identifier, record['type'],
+                                            record['name'], record['content'])
+        create_result = self._create_record(rtype, name, content)
+
+        LOGGER.debug('update_record: %s', (delete_result and create_result))
+        return delete_result and create_result
+
+    def _delete_record(self, identifier=None, rtype=None, name=None, content=None):
+        if identifier:
+            records = self._list_records(rtype)
```
adferrand: Same as above in update method, to just get all records in the zone then filter.

adferrand:
NB: Don't mind the failure in AppVeyor if it is about Python 3.4 environment, it is broken for now.

lexicon/providers/dinahosting.py
iperurena: Agreed. While I was at it I modified the _list_records to correct the lint issue and followed one of your suggestions from another PR about extracting the pseudo-id to its own _identifier method.

AnalogJ:
Hey @iperurena 
I took a look at the PR, and my feedback is pretty similar to @adferrand 
One thing I did notice is that there's alot of unsupported standard record types: `['MX', 'SRV', 'NS', 'SOA', 'LOC']`
Can you give a little bit of context around what Dinahosting does with those?

Thanks again for the PR! :tada:

iperurena:
Hey @AnalogJ 

Regarding the unsupported type list. To my knowledge, this API or the platform itself does not offer any mechanism to define LOC or SOA type records. 

As for the other types, I found out that the API either supports them only partially or the methods are somewhat designed to work as a backend to the platform's UI rather than being used programmatically. Thus, I deemed them unsuitable for the purposes of this library.

To elaborate, I only managed to add MX records using what the API calls "simplified MX". It won't let you define each records priority and it introduces a weird dependency with A type records when listing/adding the record.

iperurena:
Looking forward to your feedback @AnalogJ , @adferrand

adferrand:
OK. I think we can merge!

iperurena:
Just saw the message @adferrand, thank you for all your feedback!

PR #438, opened 2019-10-01
add RcodeZero (https://www.rcodezero.at) provider

MikeAT:
this PR implements a provider for the RcodeZero DNS API (https://my.rcodezero.at/api-doc/)


CapsuleCD:
Hi.

[CapsuleCD](https://www.capsulecd.com/project/github/AnalogJ/lexicon/pullrequests/438)

---

MikeAT:
HI,

Is there anything I can do get this PR merged?

Best,
Michael

lexicon/providers/rcodezero.py
```diff
@@ -0,0 +1,196 @@
...
+        self._zone_data = None
+        self.api_endpoint = 'https://my.rcodezero.at/api/v1'
+
+    def _authenticate(self):
+        if self._zone_data is None:
+            self._zone_data = self._get('/zones/' +self.domain)
+
+        self.domain_id = self.domain
+
+    # Create record. If record already exists with the same content, do nothing'
```
adferrand: Technically, the API is always called, even if the record already exists, but will have no effect on the existing record set. Could you update this inline comment accordingly?

lexicon/providers/rcodezero.py
```diff
@@ -0,0 +1,196 @@
...
+                                    })
+        # if the request fails for any reason, throw an error.
+        if response.status_code >= 400:
+            LOGGER.error('Bad Request: %s', response.text)
+        response.raise_for_status()
+        return response.json()
+
+    # generate a unique id for a give record
+    def _make_identifier(self, rtype, name, content):  # pylint: disable=no-self-use
+        return "{}/{}={}".format(rtype, name, content)
```
adferrand: For the sake of CLI usability, could you hash this identifier to have a short string exposed to the user as an ID? An example code can be found here: https://github.com/AnalogJ/lexicon/blob/cc985166aead697727e56517c58061693ea087ce/lexicon/providers/godaddy.py#L227-L233

lexicon/providers/rcodezero.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def _parse_identifier(self, identifier):  # pylint: disable=no-self-use
```
adferrand: If you apply my comment above, you will not be able to extract out `rtype`, `name` and `content` from the ID. Instead, you will need to call the API to get all records, then calculate on-the-fly the ID for each return record, to check if it matches with the provided ID.

I think this is the best approach, since it does not expose the internal of the provider to the user, and allow transparent future migrations if needed. This, on top of the advantage to use a simpler ID format on the CLI.

lexicon/providers/rcodezero.py
```diff
@@ -0,0 +1,196 @@
...
+        parts = identifier.split('/')
+        rtype = parts[0]
+        parts = parts[1].split('=')
+        name = parts[0]
+        content = "=".join(parts[1:])
+        return rtype, name, content
+
+    def _clean_content(self, rtype, content):
+        if rtype in ("TXT", "LOC"):
```
adferrand: There is an existing function in `BaseProvider` that can clean the entry from superfluous quotes, but it applies only to the `TXT` entries, not the `LOC`. I recognize that it is worth here an extra function for the specific usage in this provider.

tests/fixtures/cassettes/rcodezero/IntegrationTests/test_provider_authenticate_with_unmanaged_domain_should_fail.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - '46'
+      content-type:
+      - application/json
+      date:
+      - Mon, 30 Sep 2019 13:25:40 GMT
+      keep-alive:
+      - timeout=5, max=100
+      server:
+      - Apache
+      set-cookie:
```
adferrand: Maybe this is something you want to filter out from the records?

lexicon/providers/rcodezero.py
MikeAT: fixed in 45e67a5

lexicon/providers/rcodezero.py
MikeAT: fixed in e865177

lexicon/providers/rcodezero.py
MikeAT: fixed in e865177

lexicon/providers/rcodezero.py
MikeAT: Do you have any plans to extend the function in BaseProvider?

tests/fixtures/cassettes/rcodezero/IntegrationTests/test_provider_authenticate_with_unmanaged_domain_should_fail.yaml
MikeAT: Hm, I'm not getting the point. What is the problem with the response data?

adferrand: There is no problem, in fact. It is just that since this is a CSRF protection cookie, maybe it contains some sensible data, and maybe you would like to configure the cassettes filtering to remove this value from the yml files committed in the repo.

lexicon/providers/rcodezero.py
adferrand: No, not now until another provider shows up that could benefit from it. At this moment we can think about merging the logic in `BaseProvider`. For now it is better to keep it in your provider.

MikeAT: ok

tests/fixtures/cassettes/rcodezero/IntegrationTests/test_provider_authenticate_with_unmanaged_domain_should_fail.yaml
MikeAT: The cookies should not contain sensible data, but I excluded them from the cassesstes files because they are not needed for the testing.

PR #469, opened 2019-12-20
Add euserv provider

mschoettle:
This PR adds support for EUserv (https://support.euserv.com/api-doc/)

### DIFF
--- a/lexicon/providers/euserv.py
+++ b/lexicon/providers/euserv.py
@@ -102,13 +102,13 @@ class Provider(BaseProvider):
         }
 
         if name:
-            data['subdomain'] = name
+            data['subdomain'] = self._subdomain_name(name)
 
         self._add_ttl(data)
         self._add_priority(data)
 
         response = self._get('kc2_domain_dns_set', data)
-        LOGGER.debug('create_record: %s', response)
+        LOGGER.debug('create_record response: %s', response)
 
         return True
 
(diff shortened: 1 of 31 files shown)


mschoettle:
I addressed all the issues from the checks. The remaining exception in python-3.8-cover is `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9b in position 27: invalid start byte`. Is this due to a broken build script or is there anything in data that causes this?

adferrand:
Hello @mschoettle!

Sorry for the delay to respond. I will study the situation tomorrow.

Still, I am very surprise that the error is occurring on only one version of Python, and only one system. Could you relaunch a second time the CI, to check if it was not a spurious failure?

mschoettle:
@adferrand Good idea. I relaunched with a force push (couldn't find another way) and the error is gone now.

lexicon/providers/euserv.py
```diff
@@ -0,0 +1,262 @@
...
+            data = {'dns_record_id': record_id}
+
+            response = response = self._get('kc2_domain_dns_remove', data)
+            LOGGER.debug('delete_record: %s', response)
+
+        return True
+
+    # Helpers
+    # url param used as subaction
+    def _request(self, action='GET', url='/', data=None, query_params=None):
```
adferrand: I would prefer we do not mutate the contract around `_request`. Indeed this function is part of the `BaseProvider`, and is overridden by all providers, giving a global consistency.

Technically, you are always using `url=/`, and so you never need to modify it. Could you pass the subaction directly in the `query_params` instead?

mschoettle: Good point, that makes sense. I changed it.

mschoettle:
> LGTM!

Thanks! How will it be merged in? Do I have to close this PR?

adferrand:
I will merge it through our release tool, but to make a new release, by the end of today!

PR #473, opened 2020-01-04
Fix auto provider handling of parameters.

oldium:
The `ConfigResolver.with_dict()` function expects the dictionary to have nested dictionaries instead of colon-separated keys. Fix the dictionary creation by ensuring the nesting is done.

Steps to reproduce the problem:

```
lexicon auto --ovh-auth-entrypoint=ovh-eu list ovh.co.uk TXT
```

Throws error `Exception: Error, entrypoint is not defined`, which is not true.


adferrand:
I wrote both the config resolver and the auto provider, and yet I missed that...

Thanks a lot for this fix! I will include it in new release with your other PR.

PR #475, opened 2020-01-06
Fix Subreg.cz nameserver name.

oldium:
Verified with Subreg.cz support, all their language variants like `regtons.com` use nameservers `ns.gransy.com`, `ns2.gransy.com` and so on. Moreover, the API endpoints (they have more of them, for example `https://subreg.cz/wsdl`, `https://regtons.com/wsdl`) do not differentiate between the language variants (i.e. the web where the user registered the domain), so the API endpoint `https://subreg.cz/wsdl` can be used for everybody.


oldium:
Currently the domain names are: [subreg.cz](https://subreg.cz), [regtons.com](https://regtons.com), [regnames.eu](https://regnames.eu). I was thinking about how to make it clear that Subreg provider is an alias for the rest - Regtons and Regnames - write it also on the website?

oldium:
Updated PR.

Still there is a way to improve the situation - `lexicon --help` currently writes that subreg is `subreg provider`, which is not true.

adferrand:
So you think it should be named `gransy`, right ? I agree we should rename the provider if it the real one behind. But I do not want calls like `lexicon subreg ...` to fail unexpectedly, so we need a migration path.

If you are willing to follow this path, here is how I would do it:
* rebrand subreg to gransy in `lexicon/providers/subreg.py` and rename the file
* rebrand subreg to gransy in `lexicon/tests/providers/test_subreg.py` and rename the file
* rename to `gransy` the folder `tests/fixtures/cassettes/subreg`
* add optional dependency to `zeep` in `setup.py` for `gransy`, as it is done for `subreg`
* add relevant documentation of this supported provider in `README.md`

Create a proxy module `lexicon/providers/subreg.py` that essentially contains:
```
from lexicon.providers.gransy import provider_parser, Provider, NAMESERVER_DOMAINS
```
(+ good comment in it explaining why this proxy module exists)

oldium:
Yes, agree.

Additionally, I would leave subreg's `NAMESERVER_DOMAINS` empty (to omit it from auto-detection), but let gransy provider accept subreg's credentials with deprecation warning (only if gransy's credentials were not present).

adferrand:
Indeed, this is a good idea.

oldium:
Something like this? Suggestions welcome :-)

oldium:
Just a note. Because of the wrong settings of nameserver, the autodetection could never pick the subreg provider actually, so I omitted the part where gransy was able to fallback to subreg credentials - the only use case was to retain arguments compatibility with previously auto-detected subreg provider, but this could not happen. So gransy takes gransy's credentials and subreg takes subreg's credentials.

oldium:
Updated - removed subreg tests and cassettes.

PR #476, opened 2020-01-06
Decode also private domains

### LINKED ISSUE #343: Provider easyDNS is broken
During code refactoring to correct pylint issues, I discovered that the provider for easyDNS is broken.

Almost all cassettes registered include 5XX or 4XX HTTP errors when the provider is invoked. As the cassettes have been registered against the live API, it suggests most likely that the provider is not functional.

Tests should fail then, but a try/except block is incorrectly written, and hide all HTTP errors. https://github.com/AnalogJ/lexicon/blob/cb004f0bc5afb96533186fa017b2c766ecbc6f39/lexicon/providers/easydns.py#L54

This provider needs to be tested again against live API, fixed if needed, and new cassettes be submitted.

@AnalogJ, I assign you the issue as you are marked as the codeowner for this provider.

### LINKED ISSUE #461: TLDExtract Private Domains for Dynamic DNS Providers
Hello all,

I'm currently putting together the plugin for dynu.com (listed in NYI proiders).
As Dynu also acts as Dynamic DNS provider with several toplevel domains as base for the dynamic domain (`yourhost.dynu.net`, `yourhost.freeddns.org`..., also, wildcards), I had some trouble putting together the plugin.

As an example, I'm making up `mydomain.dynu.net` as my target dynamic dns hostname.

Now, the `tldextract` package used to determine the part of the domain that belongs to the toplevel will spit out `net` as the toplevel, `dynu` as the TLD and then drops `mydomain` in further processing as seen [in client.py](../blob/master/lexicon/client.py#L43).

In turn, finding the right domain from the list of dns entries in `_authenticate` is not possible by default (as `self.domain` is set to `dynu.net`).

I discovered two workarounds for this:

1. use `--delegated "mydomain.dynu.net"` to explicitly target the subdomain
2. change the code [in client.py](../blob/master/lexicon/client.py#L41) to this:

```python
extract = tldextract.TLDExtract(include_psl_private_domains=True)

# Process domain, strip subdomain
domain_parts = extract(
    self.config.resolve('lexicon:domain'))
runtime_config['domain'] = '{0}.{1}'.format(
    domain_parts.domain, domain_parts.suffix)
```

The latter is taken from [the tldextract README](https://github.com/john-kurkowski/tldextract#public-vs-private-domains).
And because Dynu probably isn't the only Dynamic DNS provider using subdomains for their users, I guess this should be the default solution.
There's a catch however that is still in ongoing development [tldextract#144](https://github.com/john-kurkowski/tldextract/pull/144):
The list of TLDs is cached on first load of the extension, so if the config is not set to `include_psl_private_domains` before the package is first initialized, it won't work. So either an update has to be triggered manually, or, lexicon should be installed and used from a virtualenv in the first place.

Since I'm already making use of method 2 in my dev enviroment, I could open a PR right away, but I'm not 100% sure on side effects for other plugins, hence my hesitation.

Thanks and best,
Chris

edit// whitespace in codeblock, typos, grammar

oldium:
Fixes #343
Fixes #461

The tldextract library ignores private domains, so domain test.us.com is wrongly extracted as subdomain=test, domain=us and suffix=com.

When allowing private domains, the extraction of test.us.com looks like subdomain=<empty>, domain=test and suffix=us.com.

Usage of `LEXICON_TLDEXTRACT_CACHE` environment variable needs to be documented somehow.

I am not sure whether this is how it should be fixed. Another workaround for private domain name resolution issue is mentioned in issue #461.

### DIFF
--- a/lexicon/client.py
+++ b/lexicon/client.py
@@ -1,6 +1,7 @@
 """Main module of Lexicon. Defines the Client class, that holds all Lexicon logic."""
 from __future__ import absolute_import
 import importlib
+import os
 
 import tldextract
 
@@ -10,6 +11,9 @@ from lexicon.config import (
     legacy_config_resolver, non_interactive_config_resolver,
 )
 
+TLDEXTRACT_CACHE_FILE_DEFAULT = os.path.join('~', '.lexicon_tld_set')
+TLDEXTRACT_CACHE_FILE = os.path.expanduser(os.environ.get("LEXICON_TLDEXTRACT_CACHE",
+                                                          TLDEXTRACT_CACHE_FILE_DEFAULT))
 
 class ProviderNotAvailableError(Exception):
     """
@@ -38,7 +42,9 @@ class Client(object):  # pylint: disable=useless-object-inheritance,too-few-publ
         runtime_config = {}
 
         # Process domain, strip subdomain
-        domain_parts = tldextract.extract(
+        domain_extractor = tldextract.TLDExtract(cache_file=TLDEXTRACT_CACHE_FILE,
+                                                 include_psl_private_domains=True)
+        domain_parts = domain_extractor(
             self.config.resolve('lexicon:domain'))
         runtime_config['domain'] = '{0}.{1}'.format(
             domain_parts.domain, domain_parts.suffix)
--- a/README.md
+++ b/README.md
@@ -233,6 +233,9 @@ Lexicon attempts to standardize authentication around the following CLI flags:
 You can see all the `--auth-*` flags for a specific service by reading the DNS service specific help: `lexicon cloudflare -h`
 
 ### Environmental Variables
+
+#### Authentication
+
 Instead of providing Authentication information via the CLI, you can also specify them via Environmental Variables.
 Every DNS service and auth flag maps to an Environmental Variable as follows: `LEXICON_{DNS Provider Name}_{Auth Type}`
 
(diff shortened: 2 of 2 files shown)


lexicon/client.py
```diff
@@ -10,6 +11,9 @@
     legacy_config_resolver, non_interactive_config_resolver,
 )
 
+TLDEXTRACT_CACHE_FILE_DEFAULT = os.path.join(os.path.dirname(__file__), '.tld_set')
```
adferrand: Two things:
* I propose to name the directory `lexicon_tld_set` to be clear about its usage
* I would like to avoid to clutter the Python installation path with cache if possible. Why not use `$(pwd)/.lexicon_tld_cache`, or `~/.lexicon_tld_cache` instead?

lexicon/client.py
```diff
@@ -10,6 +11,9 @@
...
+TLDEXTRACT_CACHE_FILE = os.path.expanduser(os.environ.get("LEXICON_TLDEXTRACT_CACHE",
```
adferrand: Good catch to invoke `os.path.expanduser` here!

oldium: That one was taken directly from tldextract 😉

lexicon/client.py
oldium: Renaming to `.lexicon_tld_set` - 👍

Regarding the file location - I think that it should be predictable (i.e. fixed), so better to me is `~/.lexicon_tld_cache` instead of current directory. One reason to have this fixed is that the file is updated only when it is old (or missing), so it could be reused from multiple instances of `lexicon` running from different directories (but not concurrently, of course, because of risk of concurrent update).

There are some recommendations for application data storage (on Linux this is `/var/cache` or `~/.cache`, on Windows - ?, on iOS - ?), but I cannot find any library handling this, so easiest is to simply use `~/.lexicon_tld_cache` and let `expanduser` function place it in the correct home directory.

Alternative would be to always download the TLDs (not store it anywhere), but I prefer the above.

If you know about a library, which could select os-dependant path for cached data (i.e. data, which can be safely deleted by the user, but which is used by the application when it exists), it should be easy to change the default.

oldium:
Updated - changed `README.md` and default path is set to `~/.lexicon_tld_set`.

PR #97, opened 2017-01-31
Add functionality for handling delegated subdomains

brianwells:
New `--delegated DELEGATED` option for specifying delegated domain.

Tested with NS1 and Python 2.7.13, but don't see any reason it wouldn't work with everything else.

PR #481, opened 2020-03-04
Add Route53 zone-id CLI parameter

theshadow27:
In DNS setups where a Route53 HostedZone is authoritative only for a subdomain (sub.example.com), and the main domain (example.com) is operated by a different provider (or even a different Route53 HostedZone on a different AWS account), the current implementation in `route53.py` fails because it is trying to match the parent domain (via `self.domain` from `BaseProvider`, which will be `example.com`, even when `another.sub.example.com` is passed to lexicon CLI) exactly to the HostedZone object's `data['Name']` which will be the real authoritative zone name (`sub.example.com`). The match fails, which results in `Exception('No domain found')`

The issue is worse when there are two Route53 HostedZones on the same AWS account (for instance to allow different IAM permissions for each of them). In this situation, the wrong HostedZone will receive the update, or (if the permissions are configured correctly) the update will fail. 

This patch addresses the issue by adding an optional `--zone-id` CLI parameter. If this parameter is present, then the provider will use the provided Zone ID (available from the AWS Console) rather than try to do a lookup and detect it. This has the added benefit of eliminating the need to add the IAM `ListHostedZones` permission to the user account performing the updates, allowing tighter security controls. When using the new parameter, only `ChangeResourceRecordSets` permission which can be restricted to a specific HostedZone by ARN need be added to the lexicon user. 

Tests pass in 2.7 and 3.7, the change is minor and does not use any python3 features. 

Thanks for the great library!


theshadow27:
Thanks for the feedback @adferrand . We were actually able to get our solution working using only the `--delegated` option from @brianwells 2017 PR (https://github.com/AnalogJ/lexicon/pull/97)  which is not in the docs ;) 

If you still think it is worthwhile to do, I am willing to record the tests using my personal AWS account when I can find some time. If we do proceed, shall I continue to use `--zone-id` or will there be a common parameter shared between google/azure/aws?

Thanks again for all the effort on this project, it has saved us SO much time.

adferrand:
Hello @theshadow27, and sorry for the delay. Quite hard to catch up with the current circumstances...

So yes, I think it is good to have a `--zone-id` parameter. For me the delegate here is more a workaround, and a zone id is more in sync with approach done on google, azure, and soon on cloudflare.

The name of the flag is ok for me. The other providers have different names, but these names are closer to the names actually used inside their respective administration consoles, so if `--zone-id` is significant for AWS, this is great.

On top of the updated integration tests, I would like to see also the method `authenticate` still do something when `--zone-id` is set. Indeed this method is supposed to validate early that the credentials provided are indeed correct to modify the DNS Zone. Could you add some call at the end of it (like get the metadata) to validate the credentials, and raise an exception if not?

Thanks in advance!

adferrand:
Hello @theshadow27, any update here?

theshadow27:
Hi @adferrand - Sorry this completely slipped between the cracks. Regarding the `authenticate` function, I have prepared a simple change to fetch the specified zone ID. This wil require read permission but only for the specific zone (rather than all zones) and thus seems to be acceptable. 

I was able to follow the CONTRIBUTING instructions as far as running the test suite goes, however upon inspecting `lexicon/tests/providers/test_route53.py` I found that the entire class is marked to skip via `@pytest.mark.skip('Cassettes needs to be regenerated')` ? So I have no way to know that the existing tests are valid at all?

TBH I am not exactly clear on the cassette/VCR concept, and what I would need to do to create new recordings. This may be a learning curve, I come from a Java background... Hopefully someone with more python experience can take over the Route53 test cases?

PR #486, opened 2020-03-25
Implement pagination for cloudflare

### LINKED ISSUE #485: Cloudflare provider only lists first DNS 100 entries for domain
How to reproduce:

* Have a CloudFlare domain with > 100 entries
* Create a new entry starting with a "high" letter, e.g. `zzzz.<domain>`
* List entries: only the first 100 entries will be returned, making it appear as though `zzzz` had not been created.

```
    def _list_records(self, rtype=None, name=None, content=None):
        filter_obj = {'per_page': 100}      ####### NOTE
        if rtype:
            filter_obj['type'] = rtype
        if name:
            filter_obj['name'] = self._full_name(name)
        if content:
            filter_obj['content'] = content

        payload = self._get(
            '/zones/{0}/dns_records'.format(self.domain_id), filter_obj)   
           ### subsequent pages not retrieved

        records = []
        for record in payload['result']:
            processed_record = {
                'type': record['type'],
                'name': record['name'],
                'ttl': record['ttl'],
                'content': record['content'],
                'id': record['id']
            }
            records.append(processed_record)

        LOGGER.debug('list_records: %s', records)
        return records
```

This causes weird errors whereby an existing entry is not returned if falls after the 100th entry.

Without specifying the "per page" parameter, only the first 20 entries are returned.

Besides a quick-n-dirty solution of raising that number to e.g. 1000, the proper solution would be to implement proper pagination support for the _get method.

marcelloromani:
Fixed #485

### DIFF
--- a/lexicon/providers/cloudflare.py
+++ b/lexicon/providers/cloudflare.py
@@ -66,7 +66,7 @@ class Provider(BaseProvider):
     # type, name and content are used to filter records.
     # If possible filter during the query, otherwise filter after response is received.
     def _list_records(self, rtype=None, name=None, content=None):
-        filter_obj = {'per_page': 100 }
+        filter_obj = {'per_page': 100}
         if rtype:
             filter_obj['type'] = rtype
         if name:
--- a/lexicon/providers/cloudflare.py
+++ b/lexicon/providers/cloudflare.py
@@ -66,7 +66,7 @@ class Provider(BaseProvider):
     # type, name and content are used to filter records.
     # If possible filter during the query, otherwise filter after response is received.
     def _list_records(self, rtype=None, name=None, content=None):
-        filter_obj = {'per_page': 100}
+        filter_obj = {'per_page': 100 }
         if rtype:
             filter_obj['type'] = rtype
         if name:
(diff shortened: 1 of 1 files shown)


lexicon/providers/cloudflare.py
```diff
@@ -74,21 +74,31 @@ def _list_records(self, rtype=None, name=None, content=None):
...
+                processed_record = {
+                    'type': record['type'],
+                    'name': record['name'],
+                    'ttl': record['ttl'],
+                    'content': record['content'],
+                    'id': record['id']
+                }
+                records.append(processed_record)
+
+            pages = payload['result_info']['total_pages']
```
adferrand, line 94: So during the first iteration of the loop, the "page" parameter is not set. Do you confirm that "page" and "pages" are still present in the response in this case, either with a paginated and non paginated result?

marcelloromani: As far as I could see, the result is always paginated.
This means `total_pages` and `page` are present even if `page=1` is not specified.

I tested the code by setting `per_page: 10` and retrieving all ~80 records from my CF zone.

marcelloromani: > On top of my inline comment, is it possible for you to register a new set of records for the tests, in order to take into account the pagination logic for fresh data from the live API?

It doesn't look like much fun to be honest :-D but I can have a look into that.

Any pointer to get me started would be great.

marcelloromani:
> Looks good, thanks a lot!
> 
> On top of my inline comment, is it possible for you to register a new set of records for the tests, in order to take into account the pagination logic for fresh data from the live API?

Sorry but I'm not quite sure how to achieve that. I guess that would mean regsitering > 100 dns entries in one of the domains I have access to and verifying the code issues a second query with `page=2` parameter?

Issue #491, opened 2020-04-11
Hetzner domain "robot" provider obsolete


l0rn:
Hetzner recently changed their HTML web based DNS management for a new one that actually has an API. Without testing it I assume that the DNS provider making use of scraping that now discontinued web form is now obsolete.

If inteded usage is for certbot though, you can use [our plugin](https://github.com/ctrlaltcoop/certbot-dns-hetzner).

I just realize it'd probably be cool to extend our Hetzner API client and merge it into this project, hm?

Hooorny:
vote for this

adferrand:
Hello @l0rn! Clearly I would appreciate a lot seeing the Hetzner provider migrated from this horrible HTML scraping to a proper API!

Indeed to construct on your plan, the API logic could be hold by Lexicon, and your plugin could re-implement the logic through the dedicated class in certbot project for lexicon-based DNS plugins (https://github.com/certbot/certbot/blob/master/certbot/certbot/plugins/dns_common_lexicon.py).

What do you think of that ? I can participate with this migration: in particular if you do not have the time to rewrite the Hetzner provider in Lexicon, I can do it, if you give me API credentials to test this out against the live Hetzner API.

l0rn:
I will take a shot in develop the Hetzner provider, gonna come back if i got questions

l0rn:
Alrighty, happy to have someone looked at #492

hetzner:
Just want to make clear a few things before you remove something you might need...

It is true that there is a public API available for DNS. It replaces the old 'mail interface' and one part of the GUI (that one at robot.hetzner.com) but not the GUI for webhosting clients (that one on konsoleh.your-server.de).

For the webhosting part you still need to scrape HTML. There might be some smaller changes at the params since last week.

l0rn:
In regards to #493 i would opt for purging hetzner web-scraping providers completely from the codebase as  they are all dysfunctional at the moment. The konsoleH provider possibly could be fixed, but I don't think the effort is worth it, as web forms tend to change, and the implementation is quite flaky.

If the maintainers agree on that I could delete the konsoleH provider in this PR, or if someone would like to fix it, it'd need to be a subject of a future PR

adferrand:
We can close this issue, since we removed the old hetzner provider.

PR #492, opened 2020-04-13
Adding a new hetzner provider compatible with dns.hetzner.com

l0rn:
Hetzner added a new DNS API while simultaniously shutting down their old
web based "Domain Robot" which was formerly implmented in lexicon
using web scraping. As far as i can see that "Domain Robot" is shut down for
good making the old Hetzner provider for this tool obsolete - BUT;
The old Hetzner provider does supply two modes, one for the obsolete robot,
one for Hetzner's konsoleH, which is their shared hosting offering, which
remains (for now?).

I moved the old web scraping based Hetzner provider to hetzner_konsoleh

~I probably would opt for deleting the really obsolete "Domain Robot" part
of the old Hetzner provider and rename it to hetzner_konsoleh.~

I would actually opt for deleting everything from the old hetzner provider, as the "robot" variant is obsolete and the konsoleh is dysfunctional because of changes to the DOM and I would argue it's unfeasable to adapt and rather would go for dropping support - but I am waiting for feedback of the maintainers.


l0rn:
This should solve #491

qqilihq:
Awesome. We’ll be really looking forward to this!

wizison:
Also looking forward to this! :)

pp0815:
With branch 09986f6 the variable PROPAGATED in **dehydrated.default.sh** throws an error. Simply removing it helps.

I think the hole special handling for PROVIDER = hetzner is obsolete now.

But at the moment i can not test it, sorry.

l0rn:
yeah, i would opt for scrapping everything related to hetzner providers relying on web scraping

l0rn:
Is there any hope we can get this looked at soon?

adferrand:
I am completely overwhelmed until tomorrow, but after this I am available. You can expect this done by the end of this week.

qqilihq:
@l0rn @adferrand I just set this up -- this works great! Thank you so much!!!

dual-oo:
@adferrand works perfect and really fast. thanks!
btw: the dehydrated example is now deprecated

level420:
> @adferrand works perfect and really fast. thanks!
> btw: the dehydrated example is now deprecated

Sorry for asking: could you please share your (of course anonymized) config for dehydrated and the new hetzner DNS? that would be great!

level420:
Here is the needed configuration for dehydrated in conjunction with the hook script from here #505 
https://github.com/AnalogJ/lexicon/pull/505#issuecomment-624614864

dagirgl:
Thank you very much for this! Works really fast! And also for/with dehydrated. But one thing I can't figure out: How do you update the "master" A-record? I tried:

`lexicon hetzner update example.com A --name='@.example.com' --content="[newIP]" --auth-token=[hetznerAuthToken]`

and

`lexicon hetzner update example.com A --name='@' --content="[newIP]" --auth-token=[hetznerAuthToken]` 

But that fails with error msg:
`requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://dns.hetzner.com/api/v1/records/[recordId]`

I also tried to escape the at-sign, but that's not the problem here, I think.
Works for `--name="www"`,  or `--name="www.example.com"` and all other record-names. What am I doing wrong? Sorry for this maybe stupid question...

PR #494, opened 2020-04-24
Added support for German DNS provider Core Networks

MasinAD:
I implemented support for the German DNS provider Core Networks. They provide a simple API with a few caveats (most importantly rate limiting logins).

My implementation passes all the tests:

```
===================== 27 passed, 2 skipped, 6 warnings in 105.94s (0:01:45) ======================
```

One major problem might be that the API requires changes to be committed which is a slow operation. That's why I decided to put the commit call in the destructor. During my tests I experienced that the destructor wasn't called anymore because Python was shutting down. Maybe there's a better solution.


MasinAD:
Okay, I don't get it. My _filter_post_data_parameters seems to get ignored as the post data is still included in the recording. Without the real login data the simulated tests always fail. With login data the simulated and live tests succeed.

OTOH, I got it running with Python 2.7.

lexicon/providers/corenetworks.py
```diff
@@ -0,0 +1,280 @@
...
+        self.domain_id = None
+        self.account_id = None
+        self.token = None
+        self.expiry = None
+        self.modified = False
+        self.auth_file = { 'token': None, 'expiry': None }
+        # Core Networks enforces a limit on the amount of logins per minute.
+        # As the token is valid for 1 hour it's sensible to store it for
+        # later usage.
+        self.auth_file_path = self._get_provider_option('auth_file') or '/tmp/corenetworks_auth.json'
```
chkpnt, line 38: I think `/var/lib/lexicon/corenetworks_auth.json` would be a more appropriate location.

lexicon/providers/corenetworks.py
```diff
@@ -0,0 +1,280 @@
...
+        LOGGER.debug("%s", str(payload))
+        self.token  = payload['token']
+        self.expiry = payload['expires'] + time.time()
+
+        # Prepare auth file and commit changes
+        self.auth_file['token']  = self.token
+        self.auth_file['expiry'] = self.expiry
+        self._commit_auth_file()
+
+    def _log(self, message):
```
chkpnt, line 272: This is probably a leftover from the development? Contains sensitive data.

lexicon/providers/corenetworks.py
```diff
@@ -0,0 +1,280 @@
...
+        if record:
+            return True
+
+        data = {
+            'name': self._relative_name(name),
+            'data': content,
+            'type': rtype
+        }
+        if self._get_lexicon_option('ttl'):
+            data['ttl'] = self._get_lexicon_option('ttl')
```
chkpnt, line 118: I just found out by accident: If the `ttl` is less than 60, this leads to a 

```
(lexicon) bash-3.2$ lexicon corenetworks --auth-username [...] --auth-password [...] create [...] --ttl 59
  File "/Users/ich/.local/share/virtualenvs/lexicon-eYduLJl4/bin/lexicon", line 11, in <module>
    load_entry_point('dns-lexicon', 'console_scripts', 'lexicon')()
  File "/Users/ich/repos/lexicon/lexicon/cli.py", line 117, in main
    results = client.execute()
  File "/Users/ich/repos/lexicon/lexicon/client.py", line 84, in execute
    return self.provider.create_record(record_type, name, content)
  File "/Users/ich/repos/lexicon/lexicon/providers/base.py", line 80, in create_record
    return self._create_record(rtype, name, content)
  File "/Users/ich/repos/lexicon/lexicon/providers/corenetworks.py", line 122, in _create_record
    payload = self._post("/dnszones/{0}/records/".format(self.domain), data)
  File "/Users/ich/repos/lexicon/lexicon/providers/base.py", line 145, in _post
    return self._request('POST', url, data=data, query_params=query_params)
  File "/Users/ich/repos/lexicon/lexicon/providers/corenetworks.py", line 209, in _request
    response.raise_for_status()
  File "/Users/ich/.local/share/virtualenvs/lexicon-eYduLJl4/lib/python2.7/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 415 Client Error: Unsupported Media Type for url: https://beta.api.core-networks.de/dnszones/[...]/records/
```

This limitation doesn't seem [to be documented](https://beta.api.core-networks.de/doc/#functon_dnszones_records_add), but maybe it's still a good idea to intercept the wrong input here.

lexicon/providers/corenetworks.py
```diff
@@ -0,0 +1,280 @@
...
+            if content:
+                record['content'] = content
+            if self._delete_record(identifier):
+                new_id = self._create_record(rtype = record['type'], name = record['name'], content = record['content'])
+                return new_id
+        else:
+            records = self._list_records( rtype=rtype, name=self._relative_name(name) )
+            if len(records) > 0:
+                for record in records:
+                    return self._update_record( record['id'], rtype, name, content )
```
chkpnt, line 152: Mh, haven't tried out, but looks like only the first match will be updated, right?

lexicon/providers/corenetworks.py
MasinAD: I thought about such directories, too, but during development I worked in a virtual environment without write permissions to `/var/lib`. Ideally, there should be some global variable like `config['data_dir']`. But I'm new to Python and new to lexicon.

lexicon/providers/corenetworks.py
MasinAD: Yes, you are right. I will fix it with the next commit.

lexicon/providers/corenetworks.py
MasinAD: Good catch. I will check for values less than 60 seconds in the next commit.

lexicon/providers/corenetworks.py
adferrand: Indeed /var/lib may not be writable, if you are not root. There is not such thing like a `data_dir` for now, since most providers just use what they get from environment variable, a given Lexicon config YAML file or CLI flags.

But here it is for caching the credentials, which is quite different. Since it contains sensitive data, I would put it in the HOME of the user, with a fallback in TMP directory if HOME does not exist. HOME can get in platform agnostic way with `os.path.expanduser("~")`, TMP similarly with `tempfile.gettempdir()` (which on Windows will be in fact a TMP directory inside the HOME user).

But the really important thing, no matter the location, is that the file has permission `0700`. Nobody expect current user should be able to see that file.

lexicon/providers/corenetworks.py
adferrand: This is a design decision here: either you update all matching records, or only the first one.

Usually we update the first one, but generate a log warning to say that more than one record was found, and only first one will be updated.

adferrand:
Hello @MasinAD, are you still willing to finish the PR?

MasinAD:
> Hello @MasinAD, are you still willing to finish the PR?

Oh, sorry, yes, I am. I will finish it at the weekend.

lexicon/providers/corenetworks.py
MasinAD: In method _commit_auth_file I assign a permission mode of 0600 as it does not need to be executable.

lexicon/providers/corenetworks.py
```diff
@@ -0,0 +1,280 @@
...
+        self.expiry = None
+        self.modified = False
+        self.auth_file = { 'token': None, 'expiry': None }
+        # Core Networks enforces a limit on the amount of logins per minute.
+        # As the token is valid for 1 hour it's sensible to store it for
+        # later usage.
+        self.auth_file_path = self._get_provider_option('auth_file') or '/tmp/corenetworks_auth.json'
+        self.api_endpoint = 'https://beta.api.core-networks.de'
+
+    def __del__(self):
```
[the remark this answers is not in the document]
MasinAD: > I see your comment on the PR about slow operations. Which delay are we talking about ?

I don't know. When asking the provider for the limits they just mentioned the login limit. The commit limit seems to be different: The API responds with a variable delay; the more often the commit function is called the slower it gets. For normal operations it's almost unnoticeable but when deleting a lot of records e.g. after running the integration tests it takes much more time to commit each single delete than to commit all deletions at once.

I will take a closer look at your proposal. As I'm new to Python I'll need a little bit more time to understand what's happening there … :smiley:

I already worked in all other comments. I guess I'll be done tomorrow evening CEST.

adferrand:
Hello @MasinAD, any update here?

adferrand:
Hello @MasinAD, quick reminder if you have the time to finish your PR :)

MasinAD:
> Hello @MasinAD, quick reminder if you have the time to finish your PR :)

Sorry for the late answer. I have to time to work on this again and I am motivated to work on this again.

lexicon/providers/corenetworks.py
```diff
@@ -0,0 +1,280 @@
...
+NAMESERVER_DOMAINS = ['core-networks.de', 'core-networks.eu', 'core-networks.com']
+
+def provider_parser(subparser):
+    """Configure provider parser for Core Networks"""
+    subparser.add_argument(
+        "--auth-username", help="specify login for authentication")
+    subparser.add_argument(
+        "--auth-password", help="specify password for authentication")
+    subparser.add_argument(
+        "--auth-file", help="specify location for authentication file")
```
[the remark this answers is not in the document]
MasinAD: Added in 16d3bd1a2bdd4725db872200ff4839a89b81ba25.

lexicon/tests/providers/test_corenetworks.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def _filter_response(self, response):
+        """See `IntegrationTests._filter_response` for more information on how
+        to filter the provider response."""
+        return response
+
+    def _test_parameters_overrides(self):
+        return {
+            'api_endpoint': 'https://beta.api.core-networks.de',
+#            'auth_username': 'apiaccount',
```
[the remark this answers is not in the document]
MasinAD: Solved in 619ebc61fec9644037942694d4da496edaf924cc

lexicon/providers/corenetworks.py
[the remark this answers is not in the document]
MasinAD: @adferrand: Okay, finally got to this point. We have decorators and contextmanager, with and yield statements, generators and iterables. Just let me say that I can wrap my head around the last two but with, yield and decorators are still beyond my scope :laughing: 

I try to slowly understand what's happening here but that's too much at once. I guess I have to live with that for now.

With those changes I run again into this error in two tests:
```
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://beta.api.core-networks.de/dnszones/fluchtkapsel.de/records/commit
```
The rest works as intended.

renne:
> Sorry for the late answer. I have to time to work on this again and I am motivated to work on this again.

@MasinAD Hi, what is the current state of the PR?

MasinAD:
> @MasinAD Hi, what is the current state of the PR?

It fails 2 tests I cannot figure out how to fix but otherwise seems to work. I updated it to latest upstream.

adferrand:
Hello @MasinAD! Thanks a lot for your update. Have you achieved the fix on the remaining tests? If so, the only thing to do to finish the PR is to update the cassettes for the integration tests, and we are good to go!

MasinAD:
I thought I updated the cassettes. Are they outdated?

renne:
@adferrand
@MasinAD
Why is the pull-request failing?

adferrand:
I think there was an issue with tox, not supporting anymore the `whitelist_externals` parameter. I just hit that one. Master branch is updated, so we should rebase the PR and see if it solves the issues.

MasinAD:
Ummm, do I have anything to do? How do I rebase the PR?

renne:
@adferrand
@MasinAD

Are there any plans to go on with this PR?

adferrand:
Well, this PR fell in the limbo given that the tests are failing and have never been fixed.

Either the original contributor can work on the issue or myself if you provide me valid credentials to actually test the provider.

renne:
@adferrand 
How can I send you a private message?

MasinAD:
If someone can tell me how to rebase the PR I can work on fixing stuff.

PR #497, opened 2020-04-27
Add Njalla provider

chapatt:
This implements Njalla's new API: https://njal.la/api/

### DIFF
--- a/lexicon/providers/njalla.py
+++ b/lexicon/providers/njalla.py
@@ -1,0 +1,146 @@
+"""Module provider for Njalla"""
+from __future__ import absolute_import
+import json
+import logging
+
+import requests
+from lexicon.providers.base import Provider as BaseProvider
+
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = [
+    '1-you.njalla.no',
+    '2-can.njalla.in',
+    '3-get.njalla.fo',
+]
+
+
+def provider_parser(subparser):
+    """Module provider for Njalla"""
+    subparser.add_argument(
+        "--auth-token", help="specify API token for authentication")
+
+class Provider(BaseProvider):
+    """Provider class for Njalla"""
+    def __init__(self, config):
+        super(Provider, self).__init__(config)
+        self.domain_id = None
+        self.api_endpoint = 'https://njal.la/api/1/'
+
+    def _authenticate(self):
+        params = {
+            'domain': self.domain,
+        }
+        result = self._api_call('get-domain', params)
+
+        if result['name'] != self.domain:
+            raise Exception('Domain not found')
+
+        self.domain_id = self.domain
+
+    # Create record. If record already exists with the same content, do nothing'
+    def _create_record(self, rtype, name, content):
+        params = {
+            'domain': self.domain,
+            'type': rtype,
+            'name': name,
+            'content': content,
+            'ttl': 10800,
+        }
+        if self._get_lexicon_option('ttl'):
+            params['ttl'] = self._get_lexicon_option('ttl')
+        result = self._api_call('add-record', params)
+
+        LOGGER.debug('create_record: %s', result)
+        return result
+
+    # List all records. Return an empty list if no records found
+    # type, name and content are used to filter records.
+    # If possible filter during the query, otherwise filter after response is received.
+    def _list_records(self, rtype=None, name=None, content=None):
+        params = {
+            'domain': self.domain,
+        }
+        result = self._api_call('list-records', params)
+
+        records = result['records']
+        processed_records = list(map(lambda record: {
+                'id': record['id'],
+                'type': record['type'],
+                'name': self._full_name(record['name']),
+                'ttl': record['ttl'],
+                'content': record['content'],
+            }, records))
+        filtered_records = list(filter(lambda record:
+            (rtype is None or record['type'] == rtype) and
+                (name is None or record['name'] == self._full_name(name)) and
+                (content is None or record['content'] == content), processed_records))
+
+        LOGGER.debug('list_records: %s', filtered_records)
+        return filtered_records
+
+    # Create or update a record.
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        if not identifier:
+            identifier = self._get_record_identifier(rtype=rtype, name=name)
+
+        params = {
+            'id': identifier,
+            'domain': self.domain,
+            'content': content,
+        }
+        result = self._api_call('edit-record', params)
+
+        LOGGER.debug('update_record: %s', result)
+        return result
+
+    # Delete an existing record.
+    # If record does not exist, do nothing.
+    def _delete_record(self, identifier=None, rtype=None, name=None, content=None):
+        if not identifier:
+            identifier = self._get_record_identifier(rtype=rtype, name=name, content=content)
+
+        params = {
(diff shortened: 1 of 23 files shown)


lexicon/providers/njalla.py
```diff
@@ -0,0 +1,145 @@
...
+            'domain': self.domain,
+            'id': identifier,
+        }
+        self._api_call('remove-record', params)
+
+        LOGGER.debug('delete_record: %s', True)
+        return True
+
+    # Helpers
+    def _api_call(self, method, params):
```
adferrand, line 113: Do you have a specific reason to not implement the method `_request` from the base provider? It seems to me that it would fit your case, and you could call `_post` in place of `_api_call`.

lexicon/providers/njalla.py
```diff
@@ -0,0 +1,145 @@
...
+            raise Exception('%d: %s' % (error['code'], error['message']))
+
+        return response['result']
+
+    def _get_record_identifier(self, rtype=None, name=None, content=None):
+        records = self._list_records(rtype=rtype, name=name, content=content)
+        if len(records) == 1:
+            return records[0]['id']
+
+        raise Exception('Unambiguous record could not be found.')
```
adferrand, line 142: Just to inform you here. You have basically two choices if more than one record is found:
* error out, like you did
* get the first matching record, and warn the user that more than a record was found, and so the first one will be taken

Most of the providers use the second option, but it is your call.

lexicon/providers/njalla.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _request(self, action='GET', url='/', data=None, query_params=None):
```
adferrand, line 144: See my comment above, about `_api_call`.

lexicon/tests/providers/test_njalla.py
```diff
@@ -0,0 +1,22 @@
...
+class NjallaProviderTests(TestCase, IntegrationTests):
+    """TestCase for Njalla"""
+    provider_name = 'njalla'
+    domain = 'example.com'
+
+    def _filter_headers(self):
+        return ['Authorization']
+
+    @pytest.fixture(autouse=True)
+    def _skip_suite(self, request):  # pylint: disable=no-self-use
```
adferrand, line 20: Are you willing to implement the advanced test suite? It is mainly about the capacity to handle multiple records for the same name (like TXTs). It is mostly useful for ACME clients, like Certbot, to issue wildcard certificates, which require two token for the same name.

It is not mandatory, but I wanted to give you an insight about it.

lexicon/providers/njalla.py
chapatt: I expected a comment on this. It just seemed weird to implement the `"request` method when the JSON-RPC API only takes POSTs. If I did, I'd still be tempted to wrap `_post` in another method that takes `method` and `params` args.

I just realized that the API docs are even require login, so some of that might not have been apparent. I'm willing to use `_request` if you feel that's important.

lexicon/providers/njalla.py
chapatt: I'm using this method in `_delete_record` so deleting the first of multiple matched records seems dangerous.

lexicon/tests/providers/test_njalla.py
chapatt: I ran into an issue with the very first one: `test_provider_when_calling_create_record_with_duplicate_records_should_be_noop`
- This is not the native behavior of Njalla (it allows duplicate records)
- I attempted doing the check myself with a `_list_records` in the beginning of `_create_record`, but it appeared to not return results for the duplicate record created only a moment before. It seems operations aren't synchronous, and it takes a while for the system to become consistent.

As completely duplicate records are permitted, record sets are also, and are returned correctly by `_list_records` with filter parameters.

I have however not implemented `def test_provider_when_calling_delete_record_with_record_set_name_remove_all`. Is deleting record sets in one go required for by certbot?

lexicon/providers/njalla.py
adferrand: I understand that for this provider specifically passing through `_request` may be a burden, but it is for a more general purpose. You see, Lexicon is composed of variety of providers developped by a lot of different people. At some point of Lexicon lifecycle, it is necessary to make some global modifications.

Having some "standard" APIs help a lot to avoid to miss something during these kind of refactoring. In particular, having the actual network requests passing through `_request` is helping to avoid missing specific logic if a global refactoring impact the network operations.

And I am completely ok to wrap the underlying `_request` calls to more suitable high level functions for the Njalla provider.

lexicon/providers/njalla.py
adferrand: I agree. My option makes more sense in update operation, but for deleting it is bad.

lexicon/tests/providers/test_njalla.py
adferrand: Alright, may be tricky to match the expected behavior for the advanced test suite. No, deleting all the record set is not mandatory: Certbot will remove a specific TXT+entry+value for each related challenge in a given certificate.

adferrand:
Really sorry for the late feedback. Appart `_request`, it is almost done! Don't mind the failure in circle ci. By the way, if you merge master, you will see it disappear.

adferrand:
Don't bother the merge from master, I did it for you.

adferrand:
Hello @chapatt!

Are you willing to finish the PR? Thanks in advance!

lexicon/providers/njalla.py
chapatt: I implemented `_request` :)

chapatt:
Sorry, I lost this PR for a bit. Please let me know if there's anything else that needs fixing up.

chapatt:
@adferrand I see I'm failing all of the automated tests now. I can't quite make sense of the errors.

After merging in master, the developer guide local test procedure doesn't seem to work anymore:
```
virtualenv -p python2.7 venv
source venv/bin/activate
py.test lexicon/tests/providers/test_njalla.py
```

adferrand:
Hello @chapatt! Really sorry for the delay, I was on vacations. Indeed I changed a little the design of integration tests.

Before you were extending your test class from `IntegrationTest`, with optional explicit markers to disable some parts of the test suite. Now the actual test suite run is defined by the class itself: `IntegrationTestsV2` or `IntegrationTestsV1`. You should use `IntegrationTestsV2` if possible, because the test suite is more complete.

So basically, replace `IntegrationTests` by `IntegrationTestsV2` in your test class, and you are good.

chapatt:
> Hello @chapatt! Really sorry for the delay, I was on vacations. Indeed I changed a little the design of integration tests.
> 
> Before you were extending your test class from `IntegrationTest`, with optional explicit markers to disable some parts of the test suite. Now the actual test suite run is defined by the class itself: `IntegrationTestsV2` or `IntegrationTestsV1`. You should use `IntegrationTestsV2` if possible, because the test suite is more complete.
> 
> So basically, replace `IntegrationTests` by `IntegrationTestsV2` in your test class, and you are good.

No worries! I updated to `IntegrationTestsV2`. I still skipped a couple tests because Njalla is not natively capable of them. If they're necessary, it will have to be done client side by querying first, then modifying behavior based on that.

It also looks like I'm failing a couple checks. I can't seem to find useful output of those builds to see why.

PR #505, opened 2020-05-06
remove hetzner special handling from dehydrated.default.sh

level420:
as it's not supported anymore


level420:
See https://github.com/AnalogJ/lexicon/pull/492

level420:
For everyone using this hook script: you need the following entries in `/etc/dehydrated/config`:
```
export PROVIDER='hetzner'
export LEXICON_HETZNER_TOKEN='your-hetzner-dns-api-token-goes-here'
```

adferrand:
Thanks a lot!

Issue #518, opened 2020-06-02
`godaddy` provider fails under v3.3.24 given FDQN


eengstrom:
Upon upgrading from 3.3.23 to .24, the `godaddy` provider fails when given a FQDN including a full stop.  e.g. `foo.example.com.` (note trailing period).  E.g.:

```
odin $ lexicon --version
lexicon 3.3.24

odin $ lexicon godaddy create example.com TXT --name=foo.example.com. --content=bar.example.com
Traceback [elided...]
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://api.godaddy.com/v1/domains/example.com/records/CNAME/foo.example.com.

odin $ pip3 install dns-lexicon==3.3.23
[elided...]
Successfully installed dns-lexicon-3.3.23

odin $ lexicon --version
lexicon 3.3.23

odin $ lexicon godaddy create example.com TXT --name=foo.example.com. --content=bar.example.com
RESULT
------
True
```

Looks like there have been a lot of changes recently, so I'm figuring this one is easy, but if @adferrand doesn't see the issue right away, LMK and I'll dive in and try to fix.

Seems like it's also related to #362, but until today I would have said that issue was OBE.

doddo:
The breaking of this one seems to correlate with the addition of workarounds for zones with CAA records (#506) (which the API does not support)

```bash
curl -s  -X GET -H "Authorization: sso-key $SSO_KEY" "https://api.godaddy.com/v1/domains/fistofsteel.xyz/records/CAA"  | jq .
{
  "code": "INVALID_VALUE_ENUM",
  "message": "type not any of: A, AAAA, CNAME, MX, NS, SOA, SRV, TXT"
}
```

The error is here in [the provider](https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/godaddy.py#L115). `name` ought to be in the short form but it is not.

        self._put('/domains/{0}/records/{1}/{2}'.format(domain, rtype, name), records)

I stumbled across some other errors too, related to updates throwing 422 expecting a list of records, but getting a single record instead running the test suite after making the above adjustments, which I found a bit confusing as they did not relate to the bits of code which I changed.

Furthermore, the API does now seem to have envolved since the godaddy provider was created, I do believe that it could be simplified somewhat.

Maybe an overhaul of the whole thing could be a good idea?

adferrand:
Hello!

[...]

PR #522, opened 2020-06-13
Revert #506

adferrand:
I have to revert #506, because of the following errors observed on the Godaddy provider:
1) `name` is used instead of `relative_name`, which leads to the error observed in #518
https://github.com/AnalogJ/lexicon/blob/ee78b98265f7ff83e47285a6441b54259a5e4870/lexicon/providers/godaddy.py#L98
https://github.com/AnalogJ/lexicon/blob/ee78b98265f7ff83e47285a6441b54259a5e4870/lexicon/providers/godaddy.py#L115
2) in the delete method, filtered records are not filtered by rtype when they will be sent to the rtype scope API endpoint `/domains/[DOMAIN]/records/[RTYPE]`, leading to `422` HTTP errors
https://github.com/AnalogJ/lexicon/blob/ee78b98265f7ff83e47285a6441b54259a5e4870/lexicon/providers/godaddy.py#L230
3) in the delete method, calls to the general API endpoint `/domains/[DOMAIN]/records` raise `422` HTTP errors for unknown reasons
https://github.com/AnalogJ/lexicon/blob/ee78b98265f7ff83e47285a6441b54259a5e4870/lexicon/providers/godaddy.py#L232

Point 1) is easily corrected. For point 2) a workaround is possible (refilter to the rtype), but in practice it still means that one cannot delete all the records for a rtype, because Godaddy refuses an empty array as a body for a given endpoint, making impossible to remove all TXT for instance with `/domains/[DOMAIN]/records/TXT`. Finally, I do not understand why 3) is happening.

Globally I could not validate the full test suite with my live account on Godaddy.

@franc6, if you are willing to submit a new request to fix the CAA problem you fixed with #506, here is all I could see as of now to explain the failure and the need to fallback.

### DIFF
--- a/lexicon/providers/godaddy.py
+++ b/lexicon/providers/godaddy.py
@@ -95,11 +95,12 @@ class Provider(BaseProvider):
         ttl = self._get_lexicon_option('ttl')
 
         # Retrieve existing data in DNS zone.
-        records = self._get('/domains/{0}/records/{1}/{2}'.format(domain, rtype, name))
+        records = self._get('/domains/{0}/records'.format(domain))
 
-        # Check if a record already matches
+        # Check if a record already matches given parameters
         for record in records:
-            if record['data'] == content:
+            if (record['type'] == rtype and self._relative_name(record['name']) == relative_name
+                    and record['data'] == content):
                 LOGGER.debug(
                     'create_record (ignored, duplicate): %s %s %s', rtype, name, content)
                 return True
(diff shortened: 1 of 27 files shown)


franc6:
Thanks for the update.  I've done a quick test, and simply removing the CAA records when uploading to /records will delete all existing CAA records. :(  It might be another week or two before I can make the adjustments above.

I didn't realize that it should be possible to delete all records of a given type, since there's no test for it.  I'll be sure to add one.  Given the way the integration tests are laid out, it might not be a good idea to just accept whatever test data I add for the other APIs, since I'm likely to assume they are working, and add test data based on that assumption.  If there's a good, easy, and quick way to avoid making such assumptions without learning how each API is used, I'd appreciate knowing that, too. :)

PR #545, opened 2020-08-26
Fixed create and update operations for GoDaddy if CAA records are present

franc6:
The GoDaddy API breaks (returning 422) if you try to submit a CAA
record.  So when you create a new record, if the domain has one or more
CAA records, the create will fail (update and delete fail the same way).
This change simply submits the new (or updated) record instead of
submitting all records.

This doesn't address the problems with delete, as the GoDaddy API still
doesn't let you delete a single record.  So for domains with CAA
records, any delete action will continue to fail.

NOTE: Unlike the previous attempt, this doesn't try to fix anything with delete; trying to work around the broken API was ugly, and wasn't working well, anyway (see #522).

### DIFF
--- a/lexicon/providers/godaddy.py
+++ b/lexicon/providers/godaddy.py
@@ -94,7 +94,7 @@ class Provider(BaseProvider):
         ttl = self._get_lexicon_option("ttl")
 
         # Retrieve existing data in DNS zone.
-        records = self._get("/domains/{0}/records".format(domain))
+        records = self._get("/domains/{0}/records/{1}/{2}".format(domain, rtype, relative_name))
 
         # Check if a record already matches given parameters
         for record in records:
@@ -115,8 +115,8 @@ class Provider(BaseProvider):
 
         records.append(data)
 
-        # Synchronize data with inserted record into DNS zone.
-        self._put("/domains/{0}/records".format(domain), records)
+        # Insert the record
+        self._put("/domains/{0}/records/{1}/{2}".format(domain, rtype, relative_name), records)
 
         LOGGER.debug("create_record: %s %s %s", rtype, name, content)
 
@@ -142,6 +142,7 @@ class Provider(BaseProvider):
         if name:
             relative_name = self._relative_name(name)
 
+        updated_record = None
         # Retrieve existing data in DNS zone.
         records = self._get("/domains/{0}/records".format(domain))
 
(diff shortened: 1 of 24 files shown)


tests/fixtures/cassettes/godaddy/IntegrationTests/test_provider_when_calling_create_record_multiple_times_should_create_record_set.yaml
```diff
@@ -162,10 +153,10 @@ interactions:
       User-Agent:
       - python-requests/2.23.0
     method: GET
-    uri: https://api.godaddy.com/v1/domains/fullm3tal.online/records
+    uri: https://api.godaddy.com/v1/domains/fullm3tal.online/records/TXT/_acme-challenge.createrecordset
   response:
     body:
-      string: '[{"data":"127.0.0.1","name":"localhost","ttl":3600,"type":"A"},{"data":"ns71.domaincontrol.com","name":"@","ttl":3600,"type":"NS"},{"data":"ns72.domaincontrol.com","name":"@","ttl":3600,"type":"NS"},{"data":"docs.example.com","name":"docs","ttl":3600,"type":"CNAME"},{"data":"_domainconnect.gd.domaincontrol.com","name":"_domainconnect","ttl":3600,"type":"CNAME"},{"data":"challengetoken1","name":"_acme-challenge.createrecordset","ttl":3600,"type":"TXT"},{"data":"challengetoken","name":"_acme-challenge.fqdn","ttl":3600,"type":"TXT"},{"data":"challengetoken","name":"_acme-challenge.full","ttl":3600,"type":"TXT"},{"data":"challengetoken","name":"_acme-challenge.test","ttl":3600,"type":"TXT"}]
+{"d      string: '[ata":"challengetoken","name":"_acme-challenge.test","ttl":3600,"type":"TXT"}]
```
adferrand, line 159: The cassette seems to be broken here, and makes the integration tests fail.

lexicon/providers/godaddy.py
```diff
@@ -94,7 +94,7 @@ def _create_record(self, rtype, name, content):
         ttl = self._get_lexicon_option("ttl")
 
         # Retrieve existing data in DNS zone.
-        records = self._get("/domains/{0}/records".format(domain))
+        records = self._get("/domains/{0}/records/{1}/{2}".format(domain, rtype, relative_name))
```
adferrand, line 97: Given that the records are filtered on their type and relative_name now, I think that the condition on line 101-105 can be drastically simplified and test only for content equality.

adferrand (changes requested):
Thanks a lot for the PR! Let fix insert and update, it will be a really good start. For delete, I think we need to reach Godaddy devs to improve their API on that matter.

tests/fixtures/cassettes/godaddy/IntegrationTests/test_provider_when_calling_create_record_multiple_times_should_create_record_set.yaml
franc6: That is strange; the typo there didn't exist when I ran the tests prior to commit.  I'm curious how it got introduced; but I can clearly see it in my local history as well as in the actual file.  This has been fixed.

lexicon/providers/godaddy.py
franc6: I agree that should be reasonable.  I left the original checks just to reduce the changes.  I've made this change as requested, too.

adferrand (approved):
Really sorry for the time it took me to validate your changes. LGTM!

PR #556, opened 2020-09-03
add support for misaka.io

ym:
Hello,

I'd like to propose adding support for misaka.io DNS service.

Regards,

### DIFF
--- a/lexicon/providers/misaka.py
+++ b/lexicon/providers/misaka.py
@@ -1,0 +1,228 @@
+"""Module provider for Misaka.IO"""
+from __future__ import absolute_import
+import json
+import logging
+
+import requests
+from lexicon.providers.base import Provider as BaseProvider
+
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = [
+    # gTLDs
+    'm1ns.com', 'm1ns.net', 'm1ns.org',
+    # new gTLDs
+    'm1ns.one', 'm1ns.moe', 'm1ns.xyz',
+    'm1ns.fyi', 'm1ns.app',
+    # ccTLDs
+    'm1ns.be', 'm1ns.io', 'm1ns.uk',
+    'm1ns.us', 'm1ns.im',
+
+    # for PTR zones only
+    'reversedns.org',
+    # legacy domains
+    'ns53.net',
+]
+
+
+def _recordset_has_record(record_set, value):
+    for record in record_set['records']:
+        if record['value'] == value:
+            return True
+    return False
+
+
+def provider_parser(subparser):
+    """Configure provider parser for Misaka.IO"""
+    subparser.add_argument(
+        "--auth-token", help="specify token for authentication")
+
+
+class Provider(BaseProvider):
+    """Provider class for Misaka.IO"""
+    def __init__(self, config):
+        super(Provider, self).__init__(config)
+        self.domain_id = None
+        self.api_endpoint = 'https://api.misaka.io/dns'
+
+    def _authenticate(self):
+        payload = self._get(f'/zones/{self.domain}/settings')
+        if not payload['id']:
+            raise Exception('No domain found')
+
+        self.domain_id = self.domain
+
+    # Create or update a record.
+    def _create_record(self, rtype, name, content):
+        name = self._relative_name(name)
+
+        identifier = f'{self.domain_id}/recordsets/{name}/{rtype}'
+        endpoint = f'/zones/{identifier}'
+        ttl = self._get_lexicon_option('ttl')
+        record = {'value': content}
+
+        existing_recordset = self._get_recordset(name, rtype)
+        if existing_recordset:
+            # append if returned recordsets doesn't include the record
+            if not _recordset_has_record(existing_recordset, content):
+                existing_recordset['records'].append(record)
+                existing_recordset['ttl'] = ttl
+                self._put(endpoint, existing_recordset)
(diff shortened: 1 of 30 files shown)


lexicon/providers/misaka.py
```diff
@@ -0,0 +1,229 @@
...
+        payload = self._get(f'/zones/{self.domain}/settings')
+        if not payload['id']:
+            raise Exception('No domain found')
+
+        self.domain_id = self.domain
+
+    # Create or update a record.
+    def _create_record(self, rtype, name, content):
+        name = self._relative_name(name)
+        print(name)
```
adferrand, line 59: I think this print is for troubleshooting purpose and should be removed.

lexicon/providers/misaka.py
```diff
@@ -0,0 +1,229 @@
...
+            if rtype and recordset['type'] != rtype:
+                continue
+
+            processed_recordset = {
+                'name': self._full_name(recordset['name']),
+                'type': recordset['type'],
+                'ttl': recordset['ttl'],
+
+                # id is useless here as we don't have record ID exposed to customers
+                'id': '{}/recordsets/{name}/{type}'.format(self.domain_id, **recordset),
```
adferrand, line 114: Nit: you are using f-strings above, since Lexicon supports only Python 3.6+. Maybe you want to take advantage of it here also ?

lexicon/providers/misaka.py
```diff
@@ -0,0 +1,229 @@
...
+    # Create or update a record.
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        # use relative_name only
+        if name:
+            name = self._relative_name(name)
+
+        new_identifier = f'{self.domain_id}/recordsets/{name}/{rtype}'
+
+        if (new_identifier == identifier or (rtype is None and name is None)):
+            # the identifier hasnt changed, or type and name are both unspecified,
```
adferrand, line 134:
```suggestion
            # the identifier hasn't changed, or type and name are both unspecified,
```

lexicon/providers/misaka.py
```diff
@@ -0,0 +1,229 @@
...
+        LOGGER.debug('list_records: %s', records)
+        return records
+
+    # Create or update a record.
+    def _update_record(self, identifier, rtype=None, name=None, content=None):
+        # use relative_name only
+        if name:
+            name = self._relative_name(name)
+
+        new_identifier = f'{self.domain_id}/recordsets/{name}/{rtype}'
```
adferrand, line 131: So misaka.io does not implement identifiers in their API, and you need to generate these IDs on Lexicon side.

Storing the parameters in the identifier to retrieve the record associated is a smart move. However, it has some drawbacks on CLI usability, since from user side you need to deal with a complex string instead of usual simple identifiers expected here (serie of number or some letters).

I personally prefer to use another approach in this situation. I calculate dynamically the identifier to use from an hash-baset approach. Typical code is:

https://github.com/AnalogJ/lexicon/blob/cc985166aead697727e56517c58061693ea087ce/lexicon/providers/godaddy.py#L227-L233

Then, when an identifier is provided to update or delete a record, I get all records of the zone, calculate the identifier for each of them, and select the one corresponding to the identifier provider.

This way, you have an identifier that is not exposing some internals of the fetching logic, and keep the identifier simple (here 8 characters).

It can be a problem however if fetching all records is quite slow for the DNS API.

What do you think of this approach?

lexicon/tests/providers/test_misaka.py
```diff
@@ -0,0 +1,20 @@
...
+class MisakaProviderTests(TestCase, IntegrationTestsV2):
+    """TestCase for Misaka.IO"""
+
+    provider_name = "misaka"
+    domain = "misaka-dns-test.stream"
+
+    def _filter_headers(self):
+        return [
+            "Authorization", "Set-Cookie",
+            "Date", "X-Misaka-Debug", "X-Request-Id", "X-Served-By", "CF-RAY", "cf-request-id",
```
adferrand, line 19: I think `Date` can be kept, right?

adferrand (changes requested):
Thanks a lot for the PR! Here is my review.

lexicon/providers/misaka.py
ym: As a comprehensive approach, we can remove `{self.domain_id}/recordsets/` part from the identifier.
Since `domain` is a mandatory argument we don't need it in identifier.

Local-calculated dynamic identifier does make it easier when typing manually on CLI but its cost is a little bit higher and I would like to avoid it if it's not required.

lexicon/tests/providers/test_misaka.py
ym: Yep, let me remove it and rebuild the fixtures.

lexicon/providers/misaka.py
adferrand: So it would leave only `{name}/{type}`. Ok I think we can work from that. I propose you the following:
* we always use the `relative_name`, not the `full_name` for the `name` parameter, since `full_name` can be processed from `relative_name` using the domain if needed.
* we encode the `{name}/{type}` in Base64URL (`base64.urlsafe_b64encode()`) to get the actual identifier for maximum portability
* as a bonus we can strip the trailing padding `=` characters with something like https://gist.github.com/cameronmaske/f520903ade824e4c30ab, but I think it not that necessary

adferrand:
Closed by https://github.com/AnalogJ/lexicon/pull/1205

PR #577, opened 2020-09-24
Google Cloud DNS | recursively paginate through zone list results

AlecTroemel:
From the [google cloud dns docs](https://cloud.google.com/dns/docs/reference/v1/managedZones/list), the manage zones list api may paginate the complete list of zones. Currently implementation is not taking this into account, so if the zone does not show up in the first pagination, we'll get a false negative and throw `Error, domain is not registered for this project` incorrectly.

This PR makes sure we paginate through all the complete zone list when searching for the zone ID.


adferrand (approved):
Perfect! Thanks a lot!

PR #602, opened 2020-10-18
Support value-domain.com

ledyba:
Hi!

I created a new provider to support [value-domain.com](https://www.value-domain.com/), a Japanese domain registry which also provide DNS server.

Could you take a look?

Thanks.


ledyba:
Hmm... Last test result causes errors, but it is not related the change (which is [just renaming variable](https://github.com/AnalogJ/lexicon/pull/602/commits/002511e43a2bc8dd975f44a2a0e1d4c0d5b59cd0)).
How can I retry the test?

lexicon/providers/valuedomain.py
```diff
@@ -0,0 +1,227 @@
...
+                continue
+            if content and processed_record["content"] != content:
+                continue
+            records.append(processed_record)
+
+        LOGGER.debug("list_records: %s", records)
+        return records
+
+    # Create or update a record.
+    def _update_record(self, identifier=None, rtype=None, name=None, content=None):
```
adferrand, line 98: You can make the function a little more flexible regarding the inputs provided.

First, as for delete action, we can check for identifier, and update the record, or try the other parameters to find the matching one(s).

Second, if an identifier is not provided, you can require only rtype and name to be set (not content). In this case, you need to handle the situation where two record would be retrieved for a given rtype+name (quite common with TXT entries):
* either you warn the user and take the first
* or you error out indicating to the user that more than one potential record to update was found.

lexicon/providers/valuedomain.py
```diff
@@ -0,0 +1,227 @@
...
+        records = []
+        full_name = None
+        if name:
+            full_name = self._full_name(name)
+
+        for record in self._get_resource_record_sets():
+            processed_record = {
+                "type": record[0].upper(),
+                "name": self._full_name(record[1]),
+                "ttl": self.ttl,
```
adferrand, line 82: Why the ttl value is not taken from the record?

lexicon/providers/valuedomain.py
```diff
@@ -0,0 +1,227 @@
...
+            LOGGER.debug("record already exists")
+            return True
+
+        resource_record_sets.append((rtype.lower(), name, self._bind_format_target(rtype, content)))
+        self._update_resource_record_sets(resource_record_sets)
+        LOGGER.debug("create_record: %s", True)
+        return True
+
+    @staticmethod
+    def _identifier(record):
```
adferrand, line 62: Note to myself. I really need to make this function generally available, since it is used in almost all providers that do not have ID native support.

adferrand (changes requested):
Thanks a lot! I propose to you some improvements for your provider.

adferrand:
Value-domain has been integrated with another PR, so I close this one.

ledyba:
@adferrand  Sorry not to respond, and thank you for reviewing!

PR #608, opened 2020-10-22
[TASK] Update Gandi API URL

megamisan:
API endpoint dns.api.gandi.net is no more working. See https://doc.livedns.gandi.net/.

Changes are:
* Endpoint URL
* How apikey is used
* Minimum supported ttl value when creating record (300)


adferrand (approved):
Thank you so much for the fix. I prepare right now a release with it.

PR #612, opened 2020-10-26
Add support for creating SSHFP records with CloudFlare

insertjokehere:
The CloudFlare API requires (not that the docs actually tell you this) that you specify the content of an SSHFP record as a dictionary via the `data` property, rather than directly via the `content` property like you do with every other record type.

## Caveats
* This change only works if you are using Lexicon as a library, as the CLI does not support the SSHFP record type

See also: https://github.com/apache/libcloud/pull/1513


adferrand (approved):
Great! Note that I have on my todo list to refactor the CLI to allow more flexible configurations for each provider.

PR #629, opened 2020-11-09
Update pyproject.toml to workaround issue with setuptools 50

### LINKED ISSUE #628: Install broken with setuptools >= 50
As of version 50 of setuptools, lexicon fails to install due to an issue in setuptools/pip (as per pypa/setuptools#2353).

Output below for reference:

```
  Running setup.py develop for dns-lexicon
    ERROR: Command errored out with exit status 1:
     command: /Users/adam/.pyvenvs/lexicon/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/Volumes/home/Projects/lexicon/setup.py'"'"'; __file__='"'"'/Volumes/home/Projects/lexicon/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps
         cwd: /Volumes/home/Projects/lexicon/
    Complete output (3 lines):
      File "<string>", line 1, in <module>
    ModuleNotFoundError: No module named 'setuptools'
    ----------------------------------------
```

It looks like the safest way to handle this at the moment is to blacklist setuptools version 50 with the assumption that it will be fixed in the next major version. I will submit a pull request for this, though it is a very minor change to fix so it may be easier to simply make the change to pyproject.toml.

admdly:
Added setuptools to requirements in pyproject.toml and blacklisted version 50 to fix #628.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [build-system]
 requires = [
+    "setuptools!=50",
     "poetry>=0.12",
 ]
 build-backend = "poetry.masonry.api"


adferrand (approved):
Thanks a lot. Indeed we encountered this issue on Certbot recently. I am glad to avoid this problem for Lexicon.

PR #637, opened 2020-11-16
fix extras declaration for "plesk" in pyproject.toml

FelixSchwarz:


FelixSchwarz:
this change is not correct - I was confused by looking at the pyproject.toml for too long.

PR #706, opened 2021-01-28
Remove mock and nose from requirements

bnavigator:
- Python >=3.3 provides mock in the stdlib unittest module. There is no need for the extra package.
- Nose is abandonded software and not used within the project.

*(edited directly in GitHub web interface, be sure to squash the commits into a more meaningful one when (if) you merge)*


adferrand (approved):
Thanks, this LGTM!

PR #739, opened 2021-02-24
Add support for Mythic Beasts API v2

lexitus:
Hi,

I've attempted to add support for Mythic Beasts' v2 DNS API. Currently it is passing the test suite, and I am using it with dehydrated to create certs, so it is somewhat functional.

I am not an experienced python developer, so apologies in advance - things I am unsure of : 

- Mythic Beasts does not provide an id for records - I am creating one using a hash.
- There is a loop in _create_record where I check to see if the record has gone live yet. This causes a lot of spam in the recording, which I don't know if it is a problem or not.
- Probably a lot of non-idiomatic things.


tests/fixtures/cassettes/mythicbeasts/IntegrationTests/test_provider_authenticate.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - '29'
+      Content-Type:
+      - application/x-www-form-urlencoded
+      User-Agent:
+      - python-requests/2.25.1
+    method: POST
+    uri: https://auth.mythic-beasts.com/login
+  response:
+    body:
+      string:
```
adferrand, line 21: Hello @lexitus, and thanks a lot for your PR!

The tests are failing because the content of the `response.body.string` parameters in the VCRpy cassettes are objects, when it is supposed to be strings. For the record, VCRpy register any HTTP exchanges during the tests in files, so they can be replayed later without the need to connect to the actual DNS provider API.

How have you registered these cassettes?

adferrand (changes requested):

tests/fixtures/cassettes/mythicbeasts/IntegrationTests/test_provider_authenticate.yaml
lexitus: Thanks for taking a look. :-)

I am trying to follow the docs for recording the cassettes. It seems I was processing the response incorrectly in _filter_response where I was removing the access token, so the response got saved wrong. The output looks more reasonable now.

lexicon/providers/mythicbeasts.py
```diff
@@ -0,0 +1,280 @@
...
+    def __init__(self, config):
+        super(Provider, self).__init__(config)
+        self.domain_id = None
+        self.api_endpoint = "https://api.mythic-beasts.com/dns/v2"
+        self.auth_token = None
+
+    def _authenticate(self):
+        payload = self._get("/zones")
+
+        if self.domain is None:
```
adferrand, line 53: In fact by construction, `self.domain` is always set by construction in the parent class (see https://github.com/AnalogJ/lexicon/blob/59419ffba7797a860251c83f73221054b3f72862/lexicon/providers/base.py#L62).

The intention in `_authenticate()` is doing the following things:
1) retrieve the credentials required to authenticate from the provider options
2) authenticate against the API and retrieve the tokens that will be used for the lifetime of the provider instance for all HTTP calls
3) verify that with these tokens we are authorized to manipulate the DNS zone targeted by `self.domain`
4) set the internal domain id on `self.domain_id` for the lifetime of the provider instance

So basically I think that here you should make the autentication token retrieval (if not set by the caller), then ensure that you can see the target zone specified by `self.domain`, then assign its value to `self.domain_id` (you do not have necessarily to hash it, it can be the domain itself).

lexicon/providers/mythicbeasts.py
```diff
@@ -0,0 +1,280 @@
...
+            ]
+        }
+        if self._get_lexicon_option("ttl"):
+            data["records"][0]["ttl"] = self._get_lexicon_option("ttl")
+
+        payload = {"success": True}
+        try:
+            payload = self._post(f"/zones/{self.domain}/records", data)
+        except requests.exceptions.HTTPError as err:
+            if err.response.status_code != 400 and err.response.json()["errors"][0][0:16] != 'Duplicate record':
```
adferrand, line 93: Maybe you could add a log debug entry here to say that the record already exists.

lexicon/providers/mythicbeasts.py
```diff
@@ -0,0 +1,280 @@
...
+                raise
+
+        if rtype == 'A' or rtype == "TXT":
+            # need to wait and poll here until verified that DNS change is live
+            changes_are_live = False
+            timeout = 300
+            start = time.time()
+
+            while(not changes_are_live and time.time() - start < timeout):
```
adferrand, line 102: Your retry loop is totally fine. But if you want to leverage some of the complexity to a built-in module, you will find examples of how to handle a retrying module with `urllib3` in dnsmadeasy and godaddy providers (in the `_request` method).

One interesting thing with urllib3 `Retry` is that you can have a loop where the wait between two attempts start with a low value that increase progressively, increasing changes that the create record will be validated faster that with a fixed wait time.

lexicon/providers/mythicbeasts.py
```diff
@@ -0,0 +1,280 @@
...
+        LOGGER.debug("payload: %s", payload)
+
+        for record in payload["records"]:
+            processed_record = {
+                "type": record["type"],
+                "name": self._full_name(record["host"]),
+                "ttl": record["ttl"],
+                "content": record["data"],
+                # no id is available, so we need to make our own
+                "id": hashlib.md5(
```
adferrand, line 144: Here also I can propose you a code to generate a "short" indentifier that is used in several modules where ids are not provided by the API: https://github.com/AnalogJ/lexicon/blob/59419ffba7797a860251c83f73221054b3f72862/lexicon/providers/azure.py#L361-L365

adferrand (changes requested):
Ok this is a great implementation! I have some propositions to align the provider behavior with the other providers in some specific points. Thanks!

tests/fixtures/cassettes/mythicbeasts/IntegrationTests/test_provider_authenticate.yaml
adferrand: Indeed, now the tests are fixed and are passing.

lexitus:
Thanks for the detailed feedback. I've tried to follow your suggestions. 

For some reason one of the tests failed - I don't really understand why. It looks like it happens during some of the setup before it gets to my code?

adferrand (changes requested):
After merging master, the tests are passing on all environments. I also took the occasion to remove most of the `409` responses in the cassettes, so that the simulated API returns `200` immediately and the tests are passing immediately instead of 40 minutes. I kept one `409` response to continue to test the retry mechanism.

As a final modification before merging, could you add Mythic Beasts in the supported providers in the README.rst file, and remove it from the potential providers in the developer guide?

lexitus:
Thanks - good idea to prune the 409s out of the cassettes 👍

I've updated the docs as requested, and also added myself to CODEOWNERS. One of the tests failed again for no obvious reason...

adferrand (approved):
I relaunched the CI and builds are passing. So LGTM!

PR #755, opened 2021-03-14
Raise exception if pdns_server and/or auth_token are not provided

rytis:
Currently if `pdns_server` is not defined we are going to fail with unclear exception (`None` has no `ends_with`). This is to add generic provider exceptin type and check for the presence of required attributes.


adferrand (approved):
Thanks, this LGTM. And I appreciate the dedicate exception, I plan in the future to build a real hierarchy of exceptions classes for Lexicon.

PR #761, opened 2021-03-17
Silent a warning in new setuptools

felixonmars:
```
/usr/lib/python3.9/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'max-line-length' will not be supported in future versions. Please use the underscore name 'max_line_length' instead
```


adferrand:
Thanks!

Issue #730, opened 2021-02-16
vultr provider has new v2 api


jlongua:
The old api v1 is depreciated and in maintenace mode.
v2 api - https://www.vultr.com/api

adferrand:
After #731 is solved, I would be glad to review a PR if you have the time to migrate the provider to the new API. I am also available to provide some advices if you need to!

PR #770, opened 2021-03-25
Implement vultr API v2

bartelsielski:
API documentation can be found here: https://www.vultr.com/api/#tag/dns

Closes https://github.com/AnalogJ/lexicon/issues/730

### DIFF
--- a/lexicon/providers/vultr.py
+++ b/lexicon/providers/vultr.py
@@ -1,6 +1,7 @@
 """Module provider for Vultr"""
 from __future__ import absolute_import
 
+import json
 import logging
 
 import requests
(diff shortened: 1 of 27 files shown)


lexicon/providers/vultr.py
```diff
@@ -23,85 +24,98 @@ class Provider(BaseProvider):
...
-        data = {
-            "domain": self.domain_id,
-            "RECORDID": identifier,
-            "ttl": self._get_lexicon_option("ttl"),
-        }
-        # if rtype:
-        #     data['type'] = rtype
+        # type cannot be specified: raise exception if it is set?
+        #if rtype:
+        #    data["type"] = rtype
```
bartelsielski, line 110: As the comment says: the record type cannot be changed using the [vultr update API][1]. So what should be done when the `rtype` variable is set?

[1]: https://www.vultr.com/api/#operation/update-dns-domain-record

adferrand (approved):
That is great. Thanks a lot!

PR #775, opened 2021-03-30
Add support for RFC2136 Dyn-DNS

trinity-1686a:
I've tested it against a local instance of [KnotDNS](https://www.knot-dns.cz/), but did not add integration tests because I was not sure how to do so without an http api to query.


lexicon/providers/ddns.py
```diff
@@ -0,0 +1,145 @@
...
+            ttl = 300
+        if not identifier:
+            rrset = self._list_records(rtype, name)
+            if len(rrset) == 1:
+                identifier = rrset[0]["id"]
+            elif len(rrset) < 1:
+                raise Exception("No records found matching type and name - won't update")
+            else:
+                raise Exception("Multiple records found matching type and name - won't update")
+        d_rtype, d_name, d_content = identifier.split(":", 2)
```
adferrand, line 112: In general for the providers where identifiers are not provided by the API, I am more in favor of an identifier generated based on a hash, typically using this kind of function: https://github.com/AnalogJ/lexicon/blob/8caaeeffc14a0fb8fb4f32fba3935213f12c5187/lexicon/providers/azure.py#L361-L367

The reasons is to get a more usable id (something short, composed only of alphanumerical characters) and more importantly, an id that cannot ship any sensible data.

However this comes at a price: you cannot extract the actual record from the id. It means that in `_update_record` and `_delete_record`, when the `identifier` is used, you will need to list all records, recalculate their identifiers and select the record matching the `identifier` provided.

Is it fine for you?

lexicon/providers/ddns.py
```diff
@@ -0,0 +1,145 @@
...
+        update.delete(d_name, d_rtype, d_content)
+        update.add(name, ttl, rtype, content)
+        self._run_query(update)
+        return True
+
+    # Delete an existing record.
+    # If record does not exist, do nothing.
+    def _delete_record(self, identifier=None, rtype=None, name=None, content=None):
+        if identifier:
+            rtype, name, content = identifier.split(":", 2)
```
adferrand, line 132: See my comment about identifiers above.

adferrand (changes requested):
Thanks a lot. On top of my inline comments, we can discuss about the testing part.

Arguably it is difficult to design the usual integrations tests, given that `vcrpy` is designed to mock HTTP over TCP calls, not other diagrams over TCP or UDP. For sure we could find another testing library to mock these calls. However at the beginning the HTTP mock approach is here to avoid to make actual calls against real production systems.

For DDNS in fact, we can go to a very different testing approach which would be:
1) Spin up a local RFC2136 compatible DNS server
2) Execute all tests for the `ddns` provider against this instance
3) Tear down the DNS server

Actually @alexzorin and I created this approach for the `rfc2136` DNS plugin in the Certbot project, using Docker (which is available in Azure Pipelines, responsible to execute tests on PRs). The relevant files are the following:
* Python logic to spin up the DNS server (bind9): https://github.com/certbot/certbot/blob/master/certbot-ci/certbot_integration_tests/utils/dns_server.py
* assets to configure the bind9 instance: https://github.com/certbot/certbot/tree/master/certbot-ci/certbot_integration_tests/assets/bind-config

If you are willing to do the work in your PR to setup this testing approach, that would be great!

However I totally understand if you do not have the time, since it is not a negligeable work. Moreover due to the nature of these tests, it is totally possible to let someone else (most likely me) make the actual implementation later. In this case I would just ask you to create the test module in `lexicon.tests.providers` with an overall `skip` directive explaining that the tests need to be written in the future.

trinity-1686a:
Your suggestion on how to test seems good, but probably too involving for me to do it myself. I'd rather let you (or someone else) do that.

adferrand (approved):
That works for me. And thanks for the cache in the identifier resolution, that is a nice move!

Issue #784, opened 2021-04-12
Provider._authenticate raises Exception


commonism:
The use of Exception without a defined str causes problems when using the providers directly.

https://github.com/certbot/certbot/blob/master/certbot/certbot/plugins/dns_common_lexicon.py#L114-L120
This does not work for infoblox, as the str mismatches.

I propose to create AuthenticateException for this purpose.

PR #786, opened 2021-04-13
introduce the AuthenticationError Exception

commonism:
Addressing #784, this introduces an Exception which is used to indicate AuthenticationError

### DIFF
--- a/lexicon/exceptions.py
+++ b/lexicon/exceptions.py
@@ -1,8 +1,20 @@
 """Lexicon exceptions module"""
 
 
-class ProviderNotAvailableError(Exception):
+class LexiconError(Exception):
+    """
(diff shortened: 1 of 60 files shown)


adferrand (changes requested):
This is an excellent idea, and I had in my bag to provide a decent hierarchy of exceptions classes for Lexicon.

To align your contribution with this design, could you move this new `AuthenticationError` class into a dedicated module, `lexicon.exceptions`, create also a `LexiconError` class and make `AuthenticationError` a subclass of it? This is for the future and to provide ability to create more dedicated errors under a unified class hierarchy.

adferrand (approved):
That's great. Thanks a lot!

Issue #568, opened 2020-09-17
Fork lexicon without poetry


ArtKoKo:
Poetry add a complexity process for build in FreeBSD
Can you keep a branch of lexicon without using poetry ?

adferrand:
In fact you can continue to use your workflow with `pip/setuptools` as before, and not care about `poetry` if you want, thanks to two mechanisms available here:
* recent versions of `pip` (`>=10.0`), implements PEP 517 and PEP 518, which is a convention to prepare in isolation the build system required in the `poetry.toml` of a project. Basically you can checkout the project, run `pip install lexicon`, and `pip` will fetch `poetry` on its own to install the project. There is a limitation for editable installations (`pip install -e`) which fall in the next point
* for older versions of `pip`, or for editable installations, an installation through `setuptools` is required and still possible. Indeed a shim for `setup.py` is been designed by the packaging community for retro-compatibility purpose, and I added it to Lexicon. In this case `setuptools` will kick in and you will not benefit from Poetry features, like advanced dependency graph resolution, but for your case I think it will be sufficient.

ArtKoKo:
Thank you but when using `setuptools`, nothing gets installed, other that a few stub files:

```
  /opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/UNKNOWN-0.0.0-py3.8.egg-info/PKG-INFO
  /opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/UNKNOWN-0.0.0-py3.8.egg-info/SOURCES.txt
  /opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/UNKNOWN-0.0.0-py3.8.egg-info/dependency_links.txt
  /opt/local/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/UNKNOWN-0.0.0-py3.8.egg-info/top_level.txt
```
I must have missed something...

adferrand:
No it is me. The shim does not cover the case of old versions of pip, just the case of `pip install -e` (editable mode). You still need at least pip 10.0+ to make it work through the build isolation.

Having a more recent version of pip on FreeBSD is not doable ?

ArtKoKo:
No because I use an another package manager and I don't want to use pip to avoid package management issues...

adferrand:
[...]

Issue #490, opened 2020-04-09
Rate limiting tests


MasinAD:
My provider does not allow so many requests to their API:
```
E           HTTPError: 429 Client Error: Too Many Requests for url: https://beta.api.core-networks.de/auth/token
```
Is there a way to slow down the tests a little bit?

adferrand:
I had this problem with two providers, GoDaddy and DNSmadeEasy, that have also these kind of rate limits. I implemented a retry mechanism directly into the provider, to retry requests after a period of time to handle reaching the ratelimit.

I think it is a little more elegant than just slowing down the tests, because it benefits also for the end users. For GoDaddy, it is a dummy "wait few seconds". For DNSmadeEasy, it is a more advanced approach with a decay factor between each try that is constructed around the known rate limit.

You can see the implementations here:
* https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/godaddy.py
* https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/dnsmadeeasy.py

PR #787, opened 2021-04-13
Use poetry-core for build

harens:
This should hopefully close https://github.com/AnalogJ/lexicon/issues/568 by switching from `poetry` to `poetry-core`. See https://github.com/python-poetry/poetry/issues/1975 for more details on the issue.

This is done as recommended by the [Poetry docs](https://python-poetry.org/docs/pyproject/#poetry-and-pep-517):

> If your pyproject.toml file still references poetry directly as a build backend, you should update it to reference poetry_core instead.

I couldn't test this on my machine as a result of https://github.com/pyca/cryptography/issues/3489 occuring with poetry. Hence I marked it as a WIP.

The only thing I'm a bit cautious of is the use of `setuptools` in the build-system. I don't think it's required for `poetry` (see https://github.com/python-poetry/poetry/issues/2956#issuecomment-698334887) but I'm not sure about `poetry-core`. However, there are other projects with a similar setup.py without needing setuptools in the pyproject.toml, such as [rich](https://github.com/willmcgugan/rich), so I thought I'd give it a shot and revert it if necessary.

See https://trac.macports.org/ticket/62682 for full discussion.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,34 +1,34 @@
 [[package]]
-name = "apipkg"
-version = "1.5"
-description = "apipkg: namespace control and lazy-import mechanism"
 category = "dev"
+description = "apipkg: namespace control and lazy-import mechanism"
+name = "apipkg"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+version = "1.5"
 
 [[package]]
-name = "appdirs"
-version = "1.4.4"
-description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
 category = "main"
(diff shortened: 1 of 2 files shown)


harens:
It seems like tests are failing for unrelated reasons relating specifically to `HTTPERROR` (maybe https://github.com/AnalogJ/lexicon/issues/490?).

commonism:
Glad to see this fails for you as well. 
I'm having the similar issues (#786), but I actually touched relevant authorization code, so … seems like something else broke.

adferrand:
Just to keep you informed that I follow the discussions, I will invest myself as soon as I have some time.

adferrand (approved):
I update the PR with master, and tests are passing. LGTM!

harens:
Thanks @adferrand.

PR #795, opened 2021-04-20
Use `domain` instead of `site_name` in plesk provider to find site id so certbot's domain guessing works

### LINKED ISSUE #794: Plesk provider assumes `site_name` equals `domain` (which is not always true)
The plesk provider assumes that the `site_name` is always the same as the `domain`: https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/plesk.py#L52

That creates problems if one wants to create a record for a subdomain (e.g. `test.example.com`) for which there is no matching plesk site. It could be the case that the record for `test.example.com` (and even all records `*.test.example.com`) are part of the site `example.com`.

Example (source for the certbot plugin [here](https://gitlab.com/spike77453/certbot-dns-plesk/)):
```
# certbot certonly --authenticator dns-plesk --dns-plesk-credentials plesk.ini -d test.example.com
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator dns-plesk, Installer None
Simulating a certificate request for test.example.com
Performing the following challenges:
dns-01 challenge for test.example.com
Cleaning up challenges
Unexpected error determining zone identifier for test.example.com: API returned error: 1013 (Site does not exist)
```

One could fix this by doing something similar to the `Client` class and strip the subdomain part https://github.com/AnalogJ/lexicon/blob/master/lexicon/client.py#L36. The `Provider` constructor could do something like:

```
        try:
            domain_extractor = tldextract.TLDExtract(
                cache_dir=_get_tldextract_cache_path(), include_psl_private_domains=True
            )
        except TypeError:
            domain_extractor = tldextract.TLDExtract(
                cache_file=_get_tldextract_cache_path(), include_psl_private_domains=True  # type: ignore
            )
        domain_parts = domain_extractor(self.domain)

        #self.site_name = self.domain
        self.site_name = f"{domain_parts.domain}.{domain_parts.suffix}"
        assert self.site_name is not None
```

And then instead of stripping the domain from `host`, one would strip the `site_name` here: https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/plesk.py#L179  
That however doesn't work any more if there actually is a site for a particular subdomain.

How about adding an optional `site-id` provider option so the user can specify the site manually in the configuration file?

spike77453:
This fixes #794

### DIFF
--- a/lexicon/providers/plesk.py
+++ b/lexicon/providers/plesk.py
@@ -49,9 +49,6 @@ class Provider(BaseProvider):
         if not self.api_endpoint.endswith(PLEX_URL_SUFFIX):
             self.api_endpoint += PLEX_URL_SUFFIX
 
-        self.site_name = self.domain
-        assert self.site_name is not None
-
         self.domain_id = None
 
         self.username = self._get_provider_option("auth_username")
@@ -110,7 +107,7 @@ class Provider(BaseProvider):
 
     def __find_site(self):
         return self.__simple_request(
-            "site", "get", {"filter": {"name": self.site_name, "dataset": {}}}
+            "site", "get", {"filter": {"name": self.domain, "dataset": {}}}
         )["result"]["id"]
 
     def _authenticate(self):


adferrand (approved):
Thanks a lot! Indeed `domain` already goes through a lot of mechanics to ensure a reliable value.

Issue #827, opened 2021-05-19
[INWX] “Object not found”-error when manipulating records


wase90168:
# Issue
I have multiple domains (zones) hosted at INWX. Some of them are registered there and some are not (registered somewhere else with the NS-Records pointing to INWX-Nameservers). INWX allows you to then create and host the zone on their nameservers.

When using lexicon to create/remove records, everything works fine for those domains registered and hosted at INWX.
For domains that are only hosted at INWX, the "Object not found" error will be returned.

I tried manipulating records directly from INWX's API and everything works fine for all domains. Requests with the domain parameter being set incorrectly will be answered by INWX with "Object not found", so I think that the issue arises from the domain parameter being wrongly transmitted by lexicon.

# Traceback
```
Traceback (most recent call last):
  File "/usr/bin/lexicon", line 11, in <module>
    load_entry_point('dns-lexicon==3.3.17', 'console_scripts', 'lexicon')()
  File "/usr/lib/python3/dist-packages/lexicon/cli.py", line 117, in main
    results = client.execute()
  File "/usr/lib/python3/dist-packages/lexicon/client.py", line 77, in execute
    self.provider.authenticate()
  File "/usr/lib/python3/dist-packages/lexicon/providers/base.py", line 69, in authenticate
    return self._authenticate()
  File "/usr/lib/python3/dist-packages/lexicon/providers/inwx.py", line 76, in _authenticate
    self._validate_response(
  File "/usr/lib/python3/dist-packages/lexicon/providers/inwx.py", line 61, in _validate_response
    raise Exception("{0}: {1} ({2})".format(
Exception: Failed to authenticate: Object does not exist (2303)
```

wase90168:
I found the issue. In the authenticate method of the INWX provider, there is the following call to check if the Domain, for which an Entry should be altered, is in the List of Domains in INWX (Line 79).
```
response = self._api.domain.info(opts)
```
But that check is against the domain list and not the nameserver list.  
According to INWX`s documentation:  
* [Documentation Domain Object:](https://www.inwx.at/en/help/apidoc/f/ch02s08.html) "The domain object provides methods for creating, deleting, listing etc. of domains."  
* [Documentation Nameserver Object: ](https://www.inwx.at/en/help/apidoc/f/ch02s08.html)"The nameserver object provides methods to manage the nameserver domains and their records."  
  
[...]

PR #828, opened 2021-05-20
[INWX] authentication - change API call from domain to nameserver (Issue #827)

wase90168:
Here the `nameserver.info` call should be used to check if a domain is registered in the INWX-Nameserver and not the domain call because it checks if the domain is bought from INWX.
According to issue #827


adferrand:
Thanks a lot! I merge it and make a new release in few hours.

PR #880, opened 2021-07-21
Fix joker domain status

cgroschupp:
the joker query-domain-list api endpoint returns multiple status.

https://joker.com/faq/content/27/20/en/query_domain_list.html


adferrand (approved):
Thanks a lot for your fix, I merge it right away and prepare a new release.

PR #886, opened 2021-07-28
filter the zone list as the return paginates at 100

KamilKeski:
record creation fails in an environment with over 100 hosted zones as the entire list is returned.  I've added passing the domain to limit the list.


KamilKeski:
nm, I see it's been done better since.

PR #909, opened 2021-08-12
Exclude integration tests from source distributions + remove MANIFEST.in

sbraz:
Hi,
Since cassettes aren't included in sdists, these tests aren't very useful. I am also removing MANIFEST.in which is unused as far as I can tell.


sbraz:
Otherwise, you could ship cassettes in PyPI tarballs. They add about 7 MiB. Too bad we can't do XZ because a complete source snapshot compressed with XZ is only 704 KiB.

adferrand:
Ok, let's clear these tests indeed. Thanks!

PR #911, opened 2021-08-13
update Namecheap's nameserver list

dudeofawesome:
add `registrar-servers.com` to Namecheap's nameserver list


adferrand:
Thank !

PR #949, opened 2021-09-15
fix conoha provider (specify auth_region when get provider option)

k-serenade:


adferrand:
Sorry for the late review. I merge it right now!

PR #954, opened 2021-09-20
Use known domain as filter

dgrothaus-ku:
When the requested zone name is known, use that zone name for API query directly instead of filtering afterwards.
API results are paginated and always checking all records of the first page may result in an error although the zone is known (but on a page past the first 100 zones)


adferrand:
Sorry for the late review. I manually updated the cassettes for the integration tests, which are passing now. Thanks!

PR #998, opened 2021-10-31
Update dreamhost to use `dns-list_records` instead of `domain-list_domains`

### LINKED ISSUE #974: Dreamhost integration will break November 2nd, 2021
It looks like Dreamhost is "retiring" some of it's API calls Nov 2nd, including `domain-list_domains` which Lexicon seems to use for validating the domain during initial auth:

https://help.dreamhost.com/hc/en-us/articles/217555767-Domain-API-commands

https://github.com/AnalogJ/lexicon/blob/db82a948febee04972bb648ac59471b292c7e394/lexicon/providers/dreamhost.py#L103

They have begin sending email to anyone using the API:

> The following API commands were run on your account in the last thirty days and will no longer function after November 2nd:
> - domain-list_domains

ryan953:
The Dreamhost api command `domain-list_domains` will be retired on Nov 2nd 2021.

This diff converts the `_authenticate()` method to use [`dns-list_records`](https://help.dreamhost.com/hc/en-us/articles/217555707-DNS-API-commands) instead. 

Fixes #974

----

The obvious difference between  `domain-list_domains` and `dns-list_records` is that list_records will return any A, CNAME, MX, TXT (and so on) record in the Dreamhost account. Potentially a lot more information compared to the list of domains.

Using the test api token (scroll to ["Test Account"](https://help.dreamhost.com/hc/en-us/articles/217560167-Application-programming-interface-overview)) we can see the difference in response size:

Before:
```
$ curl -so /dev/null "https://api.dreamhost.com/?key=6SHU5P2HLDAYECUM&cmd=domain-list_domains&format=json"  -w '%{size_download}'
752
```
After:
```
$ curl -so /dev/null "https://api.dreamhost.com/?key=6SHU5P2HLDAYECUM&cmd=dns-list_records&format=json"  -w '%{size_download}'
15940
```
This is just an example account but illustrates the difference. Any account could have a large number of domain, and/or a large number of records per domain.

I added the condition that the found record should be of type "A" or "AAAA". It's trivial for someone to add any domain (like facebook.com) to their account and create that record though. Doing so puts `facebook.com` in both the responses for both `dns-list_records` and `domains-list_domains`. I can see that behavior today in my account.
This means that `_authenticate()` wasn't, and still doesn't, do a strong ownership check of the domain name. Instead is does verify that the api token works. We shouldn't need that ownership check though, because if nameservers point somewhere else nothing will read that dns record.

--- 

The notice about all the deprecated Dreamhost API's is [here](https://help.dreamhost.com/hc/en-us/articles/217560167-Application-programming-interface-overview), but it's likely to change after Nov 2nd. Copied here for reference:
> Upcoming EOL
>
> As of November 2, 2021, the following API commands will be removed and no longer function.
> 
> Account API commands
> Domain API commands
> Mail API commands
> MySQL API commands
> Rewards API commands
> User API commands
> The following three commands will continue to function normally.
> 
> Announcement List API commands
> API metacommands
> DNS API commands

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -29,7 +29,7 @@ lexicon/providers/dnsimple.py       @analogj
 lexicon/providers/dnsmadeeasy.py    @analogj @nydr
 lexicon/providers/dnspark.py        @analogj
 lexicon/providers/dnspod.py         @analogj
-lexicon/providers/dreamhost.py      @chhsiao1981
+lexicon/providers/dreamhost.py      @chhsiao1981 @ryan953
 lexicon/providers/dynu.py           @HerrFolgreich
 lexicon/providers/easydns.py        @analogj
 lexicon/providers/easyname.py       @astzweig @rqelibari
--- a/lexicon/providers/dreamhost.py
+++ b/lexicon/providers/dreamhost.py
@@ -100,14 +100,15 @@ class Provider(BaseProvider):
 
     def _authenticate(self):
         self.domain_id = None
-        payload = self._get("domain-list_domains")
+        payload = self._get("dns-list_records")
         data = payload.get("data", None)
         if data is None:
             raise AuthenticationError("Domain not found")
 
-        for domain in data:
-            if domain.get("domain", "") == self.domain:
+        for record in data:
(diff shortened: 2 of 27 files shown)


adferrand:
Thanks, I prepare a release right now.

PR #1016, opened 2021-11-14
Update of the netcup nameserver

mustermann2021:
Source: DNS lookup

second-dns.netcup.net.
third-dns.netcup.net.
root-dns.netcup.net.


adferrand (approved):
Thanks a lot !

PR #1031, opened 2021-11-25
Ensure TTL is provided as an integer

parthjoshi-pc:
Providing the `ttl` as an environment variable (LEXICON_TTL) causes the GoDaddy API to throw a 422 error since it expects an integer but receives a string. This one-line change ensures that the `ttl` if provided is always an integer


adferrand:
Great, thanks a lot !

PR #1102, opened 2022-01-29
Add Webgo Provider

mod242:
Created a new Provider for Webgo. This Hoster doesn't expose an regular API but an Webinterface. 
My Implementation is heavily inspired by the henet provider.


mod242:
Unfortunately I cam across one Problem when trying to implement my provider with acme.sh. Want to figure out what goes wrong here and will try to correct asap. Sorry for the noise, should have looked better before opening the pull-request

mod242:
I needed to add special handling for the main A-Record of an domain since this is handled very different by webgo

adferrand (approved):
Looks good to me. Thanks a lot!

PR #1201, opened 2022-04-17
Fix Yandex provider (SRV, MX qtypes)

clickfreak:
Yandex API has changed, so i made little fix to make it worked again


adferrand (approved):
Just a little problem about typing, but I can fix that on my side. Thanks a lot!

PR #507, opened 2020-05-14
Add support for Name.com

Jamim:
Hello,

I'd like to propose adding support for the [Name.com](https://www.name.com/) domain name registrar.
Name.com API docs: https://www.name.com/api-docs/DNS

Hope you might find these changes useful.

Best regards!

PR #1212, opened 2022-04-24
Add support for Name.com

giuseongit:
Name.com API docs: https://www.name.com/api-docs/DNS

This is a follow-up of https://github.com/AnalogJ/lexicon/pull/507. I've worked an all the comments except for https://github.com/AnalogJ/lexicon/pull/507#discussion_r439760986 and https://github.com/AnalogJ/lexicon/pull/507#discussion_r439761391.
I've seen that the tests are slightly changed, so I've fixed that too.

pinging @adferrand as he was the reviewer of the previous pr

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -56,6 +56,7 @@ lexicon/providers/memset.py         @tnwhitwell
 lexicon/providers/misaka.py         @misakaio @ym
 lexicon/providers/mythicbeasts.py   @lexitus
 lexicon/providers/namecheap.py      @pschmitt @rbelnap
+lexicon/providers/namecom.py        @Jamim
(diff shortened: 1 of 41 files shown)


adferrand:
Hello @giuseongit, do you need some help to troubleshoot the test failures on Python 3.7?

giuseongit:
@adferrand yes please, I thought that changing some tests could resolve this issue, but I cannot make the tests pass

adferrand:
No big deal, the internals of `unittest.mock` have changed between Python 3.7 and 3.9. It is better to use `unitest.mock.call` to check mocked calls since the class will cover this discrepencies in term of implementation for you.

PR #1230, opened 2022-05-03
fix provider:easyname entity__name parsing

wuxxin:
current web interface presents entity__name as follows:

```html
<td class="entity__name" style="">
subdomain<span style="color: var(--color-grey-500);">.maindomain.toplevel</span>
</td>
```
this breaks parsing as string. this patch fixes the parsing.


adferrand:
Thanks a lot !

PR #1237, opened 2022-05-06
Add two part TLD fix

simonmcnair:
fix for domains like .co.uk which are two part.


adferrand:
Indeed `tldextract` is quite relevant here. It is also used at higher level in the lexicon client for similar purpose.

simonmcnair:
Sorry for my ignorance, but easiest way to test this from within dnsrobocert ?  I'm guessing the version of lexicon will need incrementing anda  build trigger ?

PR #1251, opened 2022-05-14
Refactor the OCI Provider and improve test coverage

Djelibeybi:
Refactor the auth logic to ensure the provider raises an exception
if the nominated auth type fails, instead of silently switching to
the other auth type in certain cases.

The tests now cover both API key and instance principal auth types
though the latter requires an OCI instance for live tests. Tests
also include testing that config validation works by using
syntactially correct but operationally invalid valies.

Signed-off-by: Avi Miller <avi.miller@oracle.com>


adferrand:
Great! Thanks a lot.

PR #1253, opened 2022-05-16
Add support of update without identifier to Yandex PDD

paskal:
Also, remove unneeded headers from Yandex PDD responses and improve the cleanup shell script.


adferrand:
Thanks a lot, with +1 for the shell script, I never updated it with the V2 of the integration tests.

PR #1283, opened 2022-06-12
Add support for Porkbun (closes #667)

### LINKED ISSUE #667: Add support for porkbun
https://porkbun.com/

guidopetri:
This PR adds support for Porkbun and closes #667.

Porkbun API doesn't have credentials to authenticate with, then access domains; you just access the endpoint and pass in your secret keys via the POST body. As a result, I called `_list_records` in the `_authenticate` function to determine whether the user had permissions to edit the DNS records for that domain.

I ran both the provider file and the test file through `black` but let me know if that's not desired or if a different code style is preferred (`black` was in the `poetry.lock` so I figured that you used that).

I also censored out my IP address that is returned by the `/ping` endpoint from Porkbun, and I replaced the domain name I used with `example.xyz`. Hopefully I didn't miss any other sensitive data, but I'd appreciate a heads up in case you see something.

Finally, the tests, when run against the live API, fail about halfway through with a 503 error. I think the API at some point stops responding to requests because it's either overloaded or it just doesn't let a specific user hit the API that hard. I stopped the tests halfway through and continued after a bit and they worked fine, though, so the recordings should be good.

### DIFF
--- a/lexicon/providers/porkbun.py
+++ b/lexicon/providers/porkbun.py
@@ -1,0 +1,171 @@
+import logging
+import requests
+from lexicon.exceptions import AuthenticationError
+from lexicon.providers.base import Provider as BaseProvider
+import json
+
+LOGGER = logging.getLogger(__name__)
+
+NAMESERVER_DOMAINS = ["porkbun.com"]
+
+
+def provider_parser(subparser):
+    """Return the parser for this provider"""
+    subparser.description = """
+        To authenticate with Porkbun, you need both an API key and a
+        secret API key. These can be created at porkbun.com/account/api .
+    """
+
+    subparser.add_argument("--auth-key", help="specify API key for authentication")
+    subparser.add_argument(
+        "--auth-secret", help="specify secret API key for authentication"
+    )
+
+
+class Provider(BaseProvider):
+    """Provider class for Porkbun"""
+
+    def __init__(self, config):
+        super(Provider, self).__init__(config)
+        self.api_endpoint = "https://porkbun.com/api/json/v3"
+        self._api_key = self._get_provider_option("auth_key")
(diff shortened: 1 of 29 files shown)


guidopetri:
@adferrand , @AnalogJ , this is hopefully good to merge. Looks like I can't assign reviewers but I'd have put you two :) Let me know if there's any edits you need.

guidopetri:
Hey, any chance this could get merged in soon?

lexicon/providers/porkbun.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def provider_parser(subparser):
+    """Return the parser for this provider"""
+    subparser.description = """
+        To authenticate with Porkbun, you need both an API key and a
+        secret API key. These can be created at porkbun.com/account/api .
+    """
+
+    subparser.add_argument("--api-key", help="specify API key for authentication")
```
adferrand, line 19: What about `--auth-key` and `--auth-secret` here?

adferrand (changes requested):
Thanks a lot for your contribution! I have a little suggestion for the name of the CLI flags to have more usual wording, but beside this we are good to go!

lexicon/providers/porkbun.py
guidopetri: That works for me, I had just used the wording from the Porkbun API page itself. I'll make the modification in a bit

guidopetri: Edited - I kept the "help" wording and internal wording the same, hope that's alright.

adferrand (approved):
LGTM! Thanks a lot.

guidopetri:
Thank you! I'm looking forward to using it myself in the certbot docker image you have :P

PR #1338, opened 2022-07-25
add missing exit condition after Yandex Cloud domain retrieval

paskal:
Before that change, dns_zone_id was not used, and cloud_id was used for retrieval every time.


paskal:
And again, thanks for the tests. They make the expectation of the provider super clear.

paskal:
Gentle ping @adferrand for review.

paskal:
Another gentle ping to @adferrand for review.

adferrand:
Thanks a lot. It makes a better use of the available parameters and less requests at the end. LGTM!

Issue #1389, opened 2022-09-05
Hetzner list only first 100 entries of domain


mander1000:
First, thanks for the awesome library we have been using for quite a long.

However, we don't see all entries if listing all entries of the domain. We reached over 100 entries within our domain managed by hetzner, or hetzner perhaps changed their interface.

Within the hetzner implementation, a filter exists to list max 100 entries, which paginates all entries. (https://github.com/AnalogJ/lexicon/blob/master/lexicon/providers/hetzner.py#L76)

I would expect to list all entries, not only the first 100. 

Removing the filter "per_page" will list all entries (as also documented by the hetzner docu). So I will upload a fix for that as well.

Cheers
mander

adferrand:
You are welcome !

PR #1390, opened 2022-09-05
remove Hetzner per_page filter

mander1000:
see #1389 for that


adferrand:
Thanks, after fixing the cassettes, this LGTM !

PR #1423, opened 2022-10-04
Fix upsertRecordSets call in the yandexcloud provider

paskal:
Documentation:
https://cloud.yandex.com/en/docs/dns/api-ref/DnsZone/upsertRecordSets

Someone at Yandex Cloud saw inconsistency in API paths and decided that it would be a good idea to make a backwards-incompatible change, after which all calls to `upsertRecordSets` started receiving 404s. This code change fixes that problem.

@adferrand could you please also release the change and release new version of dnsrobocert after merging it? All certificate renewals relying on yandexcloud are broken now because of that.

In the meantime, I've opened a support request with them, urging developers to fix the work of the previously advertised URL.

### DIFF
--- a/lexicon/providers/yandexcloud.py
+++ b/lexicon/providers/yandexcloud.py
@@ -245,7 +245,7 @@ class Provider(BaseProvider):
                 raise error
 
         payload = self._post(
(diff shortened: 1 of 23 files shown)


adferrand:
No problem, I do that.

PR #1435, opened 2022-10-18
Don't fail tests if nslookup is unavailable

dpifke:
Not all systems have the nslookup command installed.  Skip test_nslookup_resolution if nslookup is missing, same as we do if there's no internet connectivity.


adferrand:
Totally makes sense. Thanks !

PR #1601, opened 2023-03-13
fix missing asterisk at end of line in url

tgruenert:
in URL last asterisk will be interpreted as footnote. so it has to be quoted


adferrand:
Thanks!

PR #1605, opened 2023-05-13
README.rst

stokito:
Fix ddns provider link

the `_rfc2136` link wasn't used but the ddns was empty.
The link was for a Wiki page which is not only about the RFC2136 and may be confusing.
Instead use the direct link to RFC


adferrand:
Thanks ! Failures on pipeline are irrelevant for this PR. I merge and prepare a new release.

PR #1606, opened 2023-06-02
Updated Azure DNS REST API version

rembik:
Updated Azure DNS REST API version to fix issue with multiple records in recordset.


rembik:
@adferrand Please have a look on this PR which fixes the use of lexicon for ACME validations with one delegated CNAME containing multiple TXT records in one record set.

API version: 2018-03-01-preview:

multiple_records_2018-03-01-preview

API version: 2018-05-01:
multiple_records_2018-05-01

adferrand:
Thanks a lot. I prepare a new release.

PR #1675, opened 2023-08-06
Support for Wedos WAPI

vojtakaniok:
Hi @AnalogJ 
I created Wedos WAPI provider.
Can you please check my implementation?
Documentation for the WAPI:
https://kb.wedos.com/en/kategorie/wapi-api-interface/


vojtakaniok:
I rerun localy the live intergration test and committed it.
The pipeline fails now. Can somebody help pleas e?

lexicon/providers/wedos.py
```diff
@@ -0,0 +1,216 @@
...
+            data["ttl"] = self._get_lexicon_option("ttl")
+
+        payload = self._post(data=self._create_payload('dns-row-add', data))
+        code = payload["response"]["code"]
+        if code == 1000:
+            validation = self._validate_changes()
+            if validation:
+                return True
+            else:
+                return False
```
adferrand, line 104: I put my comment on the first occurrence of this return of `False`, but it is valid also for the other occurrences for list, update and delete.

Returning `False` may not be the best thing to do when something wrong happened during the registration of a new record. Indeed there is no strong behavior associated to this (eg. non zero exit code, or overall exception raised by the client). It is just a boolean returned in the output.

It seems to me that here your provider gives the capability to initiate a post validation of the record, and that failing the validation is a real issue (bad input, duplicates etc.).

So instead I would recommend you to create a subclass of the `LexiconError` exception, specific to your provider, like `PostValidationError` and raise it with relevant information given by the API response. This way it becomes a proper failure with non zero exit code on the cli, and it is a catcheable exception for integrators. 

Wh do you think of that?

lexicon/tests/providers/test_wedos.py
```diff
@@ -0,0 +1,48 @@
...
+
+        return request
+
+    def _filter_headers(self):
+        return ['Authorization']
+
+    def _filter_query_parameters(self):
+        return ['secret_key']
+
+    def _filter_response(self, response):
```
adferrand, line 44: I don't thing you need to override this method.

adferrand (changes requested):
Overall, very good work ! I just have two comments, including a potential better error management for validations failures.

lexicon/providers/wedos.py
vojtakaniok: Well, maybe the naming of the _validate_changes is not a good choice.
It is this command: https://kb.wedos.com/en/wapi-api-interface/wapi-command-dns-domain-commit/
It is some confirmation that all changes are done and it should be distributed to the DNS.
Without this command, the changes will be applied after 5 min.

vojtakaniok: But I will introduce the LexiconError subclass and use it when whatever create/update/delete fails

vojtakaniok: Or just throwing LexiconError will be enough?

vojtakaniok: I changed it to raising LexiconError. 
If it will be not enough ping me.

adferrand (approved):
All good for me ! Thanks a lot !

PR #1720, opened 2023-09-16
Update activation command

hatamiarash7:
We can use the `shell` command of Poetry to activate the virtual environment easily. Just forget the old ways.


adferrand:
Hello @hatamiarash7 thanks a lot for the PR. I kept the manual activation using the scripts provided by the virtual environment on purpose. Indeed there are several issues with how `poetry shell` works, in particular for Windows (https://github.com/python-poetry/poetry/issues/6721 for instance).

PR #1722, opened 2023-09-17
Add Arvancloud provider

hatamiarash7:
- Add Arvancloud provider
  - Provider file
  - Tests
  - Documentations
- Ignore `ResourceWarning` for `unclosed ssl.SSLSocket`


src/lexicon/_private/providers/arvancloud.py
```diff
@@ -0,0 +1,267 @@
...
+
+        return output
+
+    # Takes record's value and puts it into a format the lexicon supports
+    def _parse_r1c_response(self, rtype: str, input: str):
+        output = {}
+
+        output.update(
+            {
+                "A": (lambda: {"value": ", ".join([" ".join(str(value) if value is not None and value != "" else "" for value in item.values()) for item in input])}),
```
adferrand, line 260: See my discussion above about how records should be represented, but at the end I think it should simply be `"A": (lambda: {"value": input['ip']})`.

Also current implementation triggers mypy: the `input` field of this method is certainly not a string.

Finally, what about the other supported records, like `AAAA`?

tests/providers/integration_tests.py
```diff
@@ -95,6 +96,8 @@ def __init__(self):
         self.provider_module = None
 
     def setup_method(self, _):
+        warnings.filterwarnings("ignore", category=ResourceWarning,
```
adferrand, line 99: I am concerned by this filter. The test is generating alerts about unclosed sockets? Seems like a problem of resource cleaning in the provider implementation.

adferrand (changes requested):

PR #1749, opened 2023-10-05
Update cloudflare.py to consider "priority" when posting

piengeng:
failed to add MX record as 'priority' is required for MX record.


adferrand:
LGTM! Thanks.

PR #1834, opened 2023-12-17
[GoDaddy] Prevent invalid tts values and fix use subdomain

GiviMAD:
Hello, I'm using these project as part of [Nginx Proxy Manager](https://nginxproxymanager.com)  and I encountered some errors trying to run the DNS challenge for GoDaddy for a subdomain. I'm able to do it with the proposed changes. 

Hope everything is ok I'm not too in touch with python. 

Amazing project by the way :)

### DIFF
--- a/src/lexicon/_private/providers/godaddy.py
+++ b/src/lexicon/_private/providers/godaddy.py
@@ -6,11 +6,13 @@ from argparse import ArgumentParser
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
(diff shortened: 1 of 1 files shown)


GiviMAD:
I messed up with the second commit, I think now it's correct, let me know if there is something wrong there. I think the CI error is not due to my changes but I'm not sure if I'm right.

BR.

adferrand:
All good to me, and sorry for the huge delay !

PR #2069, opened 2024-10-14
Update the URI for Porkbun's API

erik434:
The new URI is live now; the old one will stop accepting API calls on December 1st, according to this email I got from them:

image

Their documentation already references the new URI and confirms that the old one is deprecated: https://porkbun.com/api/json/v3/documentation

I confirmed the new URI seems to work here by running a few manual commands (like `lexicon porkbun list <mydomain> A`), and was also able to use the updated porkbun.py to generate new Let's Encrypt certs via [dehydrated](https://github.com/dehydrated-io/dehydrated) in a live [TKL Core](https://www.turnkeylinux.org/core) VM.

As I was updating the Porkbun test cassettes, I ran into Porkbun's rate limits, so ended up adding retries when we get a 503 response, following a pattern I saw used in the DNSMadeEasy provider. This lets the tests pass, but the 503's clutter up the recorded data, so I also filtered those responses out in `_filter_response`.

Please take a look and let me know if there's anything else I should do here. Thanks!

### DIFF
--- a/src/lexicon/_private/providers/porkbun.py
+++ b/src/lexicon/_private/providers/porkbun.py
@@ -32,7 +32,7 @@ class Provider(BaseProvider):
 
     def __init__(self, config):
         super(Provider, self).__init__(config)
-        self.api_endpoint = "https://porkbun.com/api/json/v3"
+        self.api_endpoint = "https://api.porkbun.com/api/json/v3"
         self._api_key = self._get_provider_option("auth_key")
         self._secret_api_key = self._get_provider_option("auth_secret")
         self._auth_data = {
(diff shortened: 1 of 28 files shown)


adferrand:
Excellent contribution, with a good usage of vcrpy, and I appreciate the inline comments to help maintaining the tests and cassettes. Thanks a lot.

I rebased your PR in https://github.com/AnalogJ/lexicon/pull/2129 with the recent changes I did who removes in particular issues associated to Python 3.8. All tests I green, so this LGTM!

erik434:
@adferrand Thanks! Happy to help.

